PR #19, opened 2016-02-03
Merge java-stellar-base into java-stellar-sdk

bartekn:
We decided to merge java-stellar-base and java-stellar-sdk into one repository and deprecate the former. The way Java works makes it difficult to export other package classes from the “parent” package, like we do in JS to separate low-level objects from classes responsible for network connections.

**What is going to change**

* All classes representing responses from Horizon server (and their deserializers and `GsonSingleton` which aren't part of public API) are moved to `org.stellar.sdk.responses` package. 
* All `org.stellar.base.xdr` classes are moved to `org.stellar.sdk.xdr` package.
* All `org.stellar.base` classes are now in `org.stellar.sdk` package.
* `TransactionBuilderAccount` interface stays so it's easy to build a transaction using a raw `KeyPair` and sequence number in `org.stellar.sdk.Account` object but it's also possible to use response from Horizon represented by `org.stellar.sdk.responses.AccountResponse` object.
* Deprecated methods will be removed:
  * `KeyPair.fromAddress` (use `KeyPair.fromAccountId`)
  * `KeyPair.getAddress` (use `KeyPair.getAccountId`)

This PR diff is quite hard to read so I created a javadoc for the changed API. It can be found here: http://bartekn.github.io/java-stellar-sdk-proposal-doc/

**Breaking changes**

This changes will be released as version `0.1.0`. We will support `0.0.*` versions of sdk and base for unspecified amount of time (we will disclose it after merging this PR) while encouraging integrators to switch to version `0.1.0`.


myrle-krantz:
Looks good to me.

PR #25, opened 2016-02-18
hashCode method for Asset classes

### LINKED ISSUE #23: AssetTypeCreditAlphaNum and AssetTypeNative both should override hashCode.
Equal objects must have equal hashCodes.  equals is overridden in these classes; so hashCodeshould be too.

https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#equals%28java.lang.Object%29

I need this to use assets in a Set.

bartekn:
This PR also makes `AssetTypeNative`, `AssetTypeCreditAlphaNum4` and `AssetTypeCreditAlphaNum12` classes immutable.

Close #23.


myrle-krantz:
Looks good.

Issue #24, opened 2016-02-17
Some transaction failures can be ameliorated with a repeat try.  Make it possible to know when such a transaction failure has occurred.


myrle-krantz:
Failures resulting from too many requests in an hour, or a bad account sequence number can be resolved by trying again at a later point in time.  But in order to do that, I need to be able to distinguish these kinds of failures from (for example) insufficient fund failures.  Being able to read this out of the SubmitTransactionResponse would be very helpful.

PR #26, opened 2016-02-18
Add rate limiting data to responses

bartekn:
This PR includes Horizon [rate limiting](https://www.stellar.org/developers/horizon/learn/rate-limiting.html) headers in response classes:
* Added `org.stellar.sdk.responses.Response` abstract class that contains rate limitting headers.
* All response classes now extend `Response` class.
* When the client is being throttled `org.stellar.sdk.requests.TooManyRequestsException` will be thrown by `execute()` methods in request classes. This exception contains `Retry-After` header value which is the number of seconds until a new window starts.

Close https://github.com/stellar/java-stellar-sdk/issues/24


myrle-krantz:
Looks good...

PR #52, opened 2017-11-03
Removing .toString on a byte array in AllowTrustOperation Builder Class

michaeljmonte:
Executing a .toString() on a byte[] will result in a string representation of that byte[]'s memory address. This will cause the assetCode of an AllowTrustOperation that has an asset of asset type ASSET_TYPE_CREDIT_ALPHANUM12 to have an assetCode of some random address. This most certainly was not intended.

### DIFF
--- a/src/main/java/org/stellar/sdk/AllowTrustOperation.java
+++ b/src/main/java/org/stellar/sdk/AllowTrustOperation.java
@@ -89,7 +89,7 @@ public class AllowTrustOperation extends Operation {
           assetCode = new String(op.getAsset().getAssetCode4());
           break;
         case ASSET_TYPE_CREDIT_ALPHANUM12:
-          assetCode = new String(op.getAsset().getAssetCode12().toString());
+          assetCode = new String(op.getAsset().getAssetCode12());
           break;
         default:
           throw new RuntimeException("Unknown asset code");


michaeljmonte:
Thanks, I have signed the Contributor License Agreement.

Maybe I'm misunderstanding something but, I think I'm going to have to disagree with you on how it isn't a major issue, maybe it isn't a major issue in the Stellar SDK bigger picture but it is a major issue on what it is trying to do versus what it is actually doing. 

Few points:
1. The code isn't using the `String(char[] value)` constructor anywhere,
2. The line `assetCode = new String(op.getAsset().getAssetCode4());`  is using the `String(byte[] bytes) ` constructor. Which does a decode of the byte array based on system's default encoding.
3. The line `new String(op.getAsset().getAssetCode12().toString());` is using the the `String(String original)` constructor. Which basically just value copies whatever you pass into it.
4. The following example demonstrates how they will turn into completely different strings.
```
public class TestBed {

	public static void main(String[] args) {
		// TODO Auto-generated method stub
		byte[] strByteArray = (new String("Some Stellar Asset!")).getBytes();
		
		String string1 = new String(strByteArray);
		String string2 = new String(strByteArray.toString());
		
		System.out.println("String 1 is: " + string1);
		System.out.println("String 2 is: " + string2);
	}

}

String 1 is: Some Stellar Asset!
String 2 is: [B@70dea4e
```

bartekn:
Oh, I see your point now and I know why there was a misunderstanding. I read the original code like:
```java
assetCode = new String(op.getAsset().getAssetCode12()).toString();
```
but in reality it was:
```java
assetCode = new String(op.getAsset().getAssetCode12().toString());
```
(check the dot location). You are obviously 100% correct.

michaeljmonte:
Awesome! Thanks for the merge!

PR #54, opened 2017-11-08
Fix memo decoding in TransactionDeserializer and add it to PageDeserializer

### LINKED ISSUE #53: Transactions deserialized in a page do not contain memo data.
Basically when deserializing a page of transactions, the memo fields for all of the transactions will be null. This is easily reproducible using the `TransactionPageDeserializerTest.testDeserialize` method.

We tracked this down to the `PageDeserializer `not registering the `TransactionDeserializer`. I would have put in a pull request for it, however there are two issues.

1. This might have been by desgin.
2. If we register `TransactionDeserializer `in `PageDeserializer `, `TransactionPageDeserializerTest.testDeserialize` will fail because of the [JSON: Memo: An empty "memo" is not being generated for a transaction even if "memo_type" is "text"](https://github.com/stellar/go/issues/149) issue I added over in the Horizon code base.

bartekn:
Fix #53

### DIFF
--- a/src/main/java/org/stellar/sdk/responses/TransactionDeserializer.java
+++ b/src/main/java/org/stellar/sdk/responses/TransactionDeserializer.java
@@ -28,17 +28,28 @@ public class TransactionDeserializer implements JsonDeserializer<TransactionResp
     if (memoType.equals("none")) {
       memo = Memo.none();
     } else {
-      String memoValue = json.getAsJsonObject().get("memo").getAsString();
+      // Because of the way "encoding/json" works on structs in Go, if transaction
+      // has an empty `memo_text` value, the `memo` field won't be present in a JSON
+      // representation of a transaction. That's why we need to handle a special case
+      // here.
       if (memoType.equals("text")) {
-        memo = Memo.text(memoValue);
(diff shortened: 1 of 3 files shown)


src/main/java/org/stellar/sdk/responses/TransactionDeserializer.java
```diff
@@ -28,17 +28,28 @@ public TransactionResponse deserialize(JsonElement json, Type typeOfT, JsonDeser
     if (memoType.equals("none")) {
       memo = Memo.none();
     } else {
-      String memoValue = json.getAsJsonObject().get("memo").getAsString();
+      // Because of the way "encoding/json" works on structs in Go, if transaction
```
nullstyle: Could you elaborate more on why java code requires changes due to go behavior?  This strikes me as a bug in the behavior in our go implementation.

bartekn: https://github.com/stellar/go/issues/149

bartekn:
Merging now because https://github.com/stellar/go/issues/149 is still not fixed.

PR #56, opened 2017-11-23
Add slim jar to build, give the fat jar the 'all' classifier

ms-novatti:


overcat:
Implemented in https://github.com/stellar/java-stellar-sdk/pull/528, thanks for your contribution.

PR #60, opened 2017-12-28
Made the AccountTest::testNullArguments() test more explicit

dexterlohnes:
Hi all.

I was reading over the tests in order to get a better idea of where to start, and while this test was written in as few lines of code as possible, for a newcomer to the project(s) like me it pays off to have things be a little more explicit when reading tests as a form of documentation in and of themselves.

As such, I've provided some minor changes to a single test in order to better reveal its intent to someone like me who isn't yet familiar with the codebase.


overcat:
Thank you for your contribution. I think it would be better to use different function names to distinguish the purposes. However, since a considerable amount of time has passed, I have decided to close this PR for now. If you are interested in updating it, please reopen it.

PR #69, opened 2018-01-17
Code to extract a Transaction instance from a (possibly base64-encoded) XDR envelope

reverbel:
Obtaining a Transaction instance from a base64-encoded XDR envelope is useful for a program that (1) somehow obtains a string representing a signed but not yet submitted transaction, (2) lets the user check if the transaction data is correct, and (3) sends the transaction to the Stellar network. The program may be a mobile app that gets the transaction string by scanning a QR code produced by a cold storage wallet.

I changed the TimeBounds class because the Stellar Laboratory generates transactions with minTIme==maxTime==0. Without the TimeBounds change, extracting such a transaction from an envelope would generate an IllegalArgumentException.

### DIFF
--- a/src/main/java/org/stellar/sdk/TimeBounds.java
+++ b/src/main/java/org/stellar/sdk/TimeBounds.java
@@ -15,8 +15,8 @@ final public class TimeBounds {
 	 * @param maxTime 64bit Unix timestamp
 	 */
 	public TimeBounds(long minTime, long maxTime) {
-		if(minTime >= maxTime) {
-			throw new IllegalArgumentException("minTime must be >= maxTime");
+		if (minTime > maxTime) {
+			throw new IllegalArgumentException("minTime must be <= maxTime");
 		}
 		
 		mMinTime = minTime;
--- a/src/main/java/org/stellar/sdk/Transaction.java
+++ b/src/main/java/org/stellar/sdk/Transaction.java
@@ -4,8 +4,11 @@ import org.apache.commons.codec.binary.Base64;
 import org.stellar.sdk.xdr.DecoratedSignature;
 import org.stellar.sdk.xdr.EnvelopeType;
 import org.stellar.sdk.xdr.SignatureHint;
+import org.stellar.sdk.xdr.TransactionEnvelope;
+import org.stellar.sdk.xdr.XdrDataInputStream;
 import org.stellar.sdk.xdr.XdrDataOutputStream;
 
+import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.nio.ByteBuffer;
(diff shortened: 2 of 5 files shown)


src/main/java/org/stellar/sdk/TimeBounds.java
```diff
@@ -15,8 +15,8 @@
 	 * @param maxTime 64bit Unix timestamp
 	 */
 	public TimeBounds(long minTime, long maxTime) {
-		if(minTime >= maxTime) {
```
bartekn: I think we should leave `=` here. In 99% cases it's an error. You have almost no cache to predict ledger close time to seconds.

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -195,17 +198,81 @@ public int getFee() {
...
 
+  /**
+   * Decodes a base64-encoded XDR {@link TransactionEnvelope}.
+   *
+   * @bug Does not work on Android, as it performs a call to the method
+   *      decode() of class org.apache.commons.codec.binary.Base64, which
+   *      conflicts with a legacy Apache class included in Android.
+   *      To decode a base64-encoded {@link TransactionEnvelope} on Android,
+   *      use {@code android.util.Base64.decode()}, then pass the result to
+   *      the method {@link #decodeXdrEnvelope(byte[])}.
```
bartekn: Please use `com.google.common.io.BaseEncoding`.

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -195,17 +198,81 @@ public int getFee() {
...
+   *
+   * @param encodedXdrTxEnvelope the base64-encoding of an XDR-encoded {@link TransactionEnvelope}.
+   * @return the decoded {@link TransactionEnvelope}
+   */
+  public static TransactionEnvelope decodeBase64XdrEnvelope(String encodedXdrTxEnvelope) {
```
bartekn: What do you think about changing this method name to `decodeBase64XdrEnvelope` (basically method overloading).

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -195,17 +198,81 @@ public int getFee() {
...
+    org.stellar.sdk.xdr.Transaction xdrTx = txEnv.getTx();
+    KeyPair sourceAccount = KeyPair.fromXdrPublicKey(xdrTx.getSourceAccount().getAccountID());
+    long sequenceNumber = xdrTx.getSeqNum().getSequenceNumber().getUint64();
+    org.stellar.sdk.xdr.Operation[] xdrOps = xdrTx.getOperations();
+    Operation[] operations = new Operation[xdrOps.length];
+    for (int i = 0; i < xdrOps.length; i++) {
+      operations[i] = Operation.fromXdr(xdrOps[i]);
+    }
+    Memo memo = Memo.fromXdr(xdrTx.getMemo());
+    org.stellar.sdk.xdr.TimeBounds xdrTimeBounds = xdrTx.getTimeBounds();
```
bartekn: Maybe we can have `fromXdr` on `TimeBounds` too (like `Memo`, `Operation`)?

src/test/java/org/stellar/sdk/TransactionTest.java
```diff
@@ -108,6 +108,20 @@ public void testBuilderSuccessPublic() throws FormatException {
             transaction.toEnvelopeXdrBase64());
   }
 
+  @Test
+  public void testTransactionFromEnvelopeXdrBase64() {
+    String xdrBase64Envelope;
+    Transaction tx;
+
+    xdrBase64Envelope = "AAAAAF7FIiDToW1fOYUFBC0dmyufJbFTOa2GQESGz+S2h5ViAAAAZAAKVaMAAAABAAAAAAAAAAAAAAABAAAAAAAAAAAAAAAA7eBSYbzcL5UKo7oXO24y1ckX+XuCtkDsyNHOp1n1bxAAAAAEqBfIAAAAAAAAAAABtoeVYgAAAEDzfR5PgRFim5Wdvq9ImdZNWGBxBWwYkQPa9l5iiBdtPLzAZv6qj+iOfSrqinsoF0XrLkwdIcZQVtp3VRHhRoUE";
```
bartekn: Would you mind constructing the transaction here instead of using XDR? It's easier to check quickly if it tests ex. time bounds.

src/main/java/org/stellar/sdk/TimeBounds.java
reverbel: I removed the = just to avoid throwing the exception when the time bounds are both unset (when `minTime` and `maxTime` both equal to 0). Perhaps we should change the if condition to
```java
    if (minTime != 0 && maxTime != 0 && minTime >=  maxTime)  {
        throw new IllegalArgumentException("minTime must be < maxTime");
    }
```
The current code also throws the exception when `minTime` is set and `maxTime` is unset. The proposal above allows setting just one of the time bounds. (I am not sure if we want to allow this.)

src/main/java/org/stellar/sdk/Transaction.java
reverbel: Will do.

reverbel: I am also removing the whole `@bug` comment, which is now obsolete.

src/main/java/org/stellar/sdk/Transaction.java
reverbel: I think method overloading is a good idea here. But you probably mean naming both methods `decodeXdrEnvelope`, right?

src/main/java/org/stellar/sdk/Transaction.java
reverbel: Sure, will do.

src/test/java/org/stellar/sdk/TransactionTest.java
reverbel: Will do.

src/main/java/org/stellar/sdk/Transaction.java
bartekn: Yeah, you're right, exactly.

src/main/java/org/stellar/sdk/TimeBounds.java
bartekn: Thanks for mentioning this. Your condition won't throw an exception when both `minTime` and `maxTime` are set and `minTime >= maxTime` We should probably have something like:
```java
// If maxTime set compare to minTime
if (maxTime != 0 && minTime >= maxTime) {
  throw new IllegalArgumentException("minTime must be < maxTime");
}
```
If `maxTime == 0`, any `minTime` value is correct.

reverbel: I agree with your conclusion (yes, we should change the condition as you indicated), but disagree with what you said about my condition. In fact, the proposed code fragment says precisely "throw an exception when both `minTime` and `maxTime` are set and `minTime >= maxTime`"! (I guess you're doing too much multitasking... :-)

Here we have an additional (and unstated) assumption: neither `minTime` nor `maxTime` can be negative. Under this assumption, the condition
```java
(minTime != 0 && maxTime != 0 && minTime >= maxTime)
```
is equivalent to 
```java
(minTime > 0 && maxTime > 0 && minTime >= maxTime)
```
Since the last two clauses above (`maxTime > 0 && minTime >=  maxTime`) imply the first clause (`minTime > 0`), the first clause is redundant and should be dropped. This is the reason your proposal is better.

Perhaps we should also check that `minTime` and `maxTime` are indeed non-negative?
```java
if (minTime < 0 || maxTime < 0) {
    throw new IllegalArgumentException("neither minTime nor maxTime can be negative");
}
else if (maxTime != 0 && minTime >= maxTime) {
    throw new IllegalArgumentException("minTime must be < maxTime");
}
```

src/test/java/org/stellar/sdk/TransactionTest.java
reverbel: @bartekn: I've made the changes your requested and need advice on how to proceed. 

I would like to update the files in this PR, but am not sure if that is possible now. The PR predated the `android` branch. Now I am working on the `android` branch (I have pulled this branch and merged my code into a local copy of the `android` branch), so the last changes are based upon the `android` branch. I guess I need to wait until `android` is merged into `master`, right?

src/main/java/org/stellar/sdk/TimeBounds.java
bartekn: > I agree with your conclusion (yes, we should change the condition as you indicated), but disagree with what you said about my condition.

Yeah, please ignore my comment, I wrote it after 6am call so I wasn't thinking clearly 😉 . However, we should make sure that we check for the case when `minTime == maxTime` and I agree we should also throw an exception when `minTime < 0 || maxTime < 0`.

src/test/java/org/stellar/sdk/TransactionTest.java
bartekn: Can you update your master branch (the branch used in this PR) with the changes? I don't think working on a PR in 2 branches is a good idea because then we'll have to wait for one branch to be merged before another.

reverbel:
Done, my master branch is updated with the latest changes.

On Mon, Feb 12, 2018 at 2:55 PM, Bartek Nowotarski <notifications@github.com

reverbel:
@bartekn: The changes you requested are in my master branch. Your recent updates (up to the 0.2.0 tag) were already merged into that branch. (The commit history got a bit ugly, though.)

Please let me know if you see any issues. Thanks!

overcat:
Thank you for your contribution, but this PR is too old and outdated, so I have decided to close it.

PR #70, opened 2018-01-19
resolves #66 preserve type information for generic response types

### LINKED ISSUE #66: Page.getNextPage returns an empty Page object
Calling `Page.getNextPage` returns an empty `Page` object when even when there is more content.  It looks like the `getNextPage` method doesn't have enough type information to construct the correct `TypeToken` (the type `T` has been erased by this point, so the `TypeToken` will just be `TypeToken<Page>` rather than e.g. `TypeToken<Page<OperationResponse>>`).

The first `Page` response is request with the correct concrete type, and the `TypeToken` is available in the `ResponseHandler.handleResponse` method.  I've put together a branch with a quick-fix for this that preserves the input `TypeToken` on any object implementing a new `TypedResponse` interface.  This lets it be re-used within `getNextPage`.

You can see it at https://github.com/davejohnclark/java-stellar-sdk/commit/7960f78c603f2726c814a96f6a1e305160d15b40  if this looks reasonable I'm happy to tidy this up and get a PR raised.

davejohnclark:
Fixes the `Page.getNextPage` method by using a new `TypedResponse` interface to allow the `ResponseHandler` to preserve the `TypeToken` on the original response `Page`, so it is present for subsequent page requests.

Also includes an update to the typing of `ResponseHandler`, which was missing the type information on the implemented `org.apache.http.client.ResponseHandler`.  This removes the need for the casts being done wherever `ResponseHandler` is used.

Added some common eclipse ignore settings.

This PR gets `Page.getNextPage` into a functional state but I see two main improvements that could be made here:

1. It doesn't fix the underlying issue that `Page` objects are being requested in different ways from different places.  It's easy for one of the ways to be broken compared to the others while this is the case.  If `Page` building were factored out somewhere common it would make sure that the same mechanism was used each time.
2. It's untested.  The current Response object tests cover deserialization, but that's not the issue here.  To cover this case would require a slightly broader test that included covering the behaviour we expect from `ResponseHandler`.  At the moment this would be difficult to achieve due to the static http utility used.  Addressing point 1. above would likely make this easier as the bulk of the logic would be moved.  As long as that class were covered the overall test coverage would be more comprehensive.

A further improvement I'd like to raise (on the topic of the static http utility being used fairly extensively) is that while it simplifies things initially I generally find the being able to control the http client configuration for sdk that I'm using ends up being a must.  There's a readily available `httpClient` instance in the `Server`, and allowing that to be provided in a constructor would give the user of the sdk full control over the client being used.  
My suggestion would involve then propagating this instance through the various request builders, so the same httpClient is used throughout, and easily injectable in one place.  I can see there are some static request methods on some of the builders, so some thought would need to go into what would happen with them.  And it is a pretty extensive change to the current design.  Ultimately I think users of the sdk would thank you when they come to needing to apply non-default configuration to the http client used.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,11 @@
 .idea/
 /local.properties
 
+# Eclipse project files
+.settings
+.classpath
+.project 
+
 # Gradle
 .gradle
 build
--- a/src/main/java/org/stellar/sdk/requests/ResponseHandler.java
+++ b/src/main/java/org/stellar/sdk/requests/ResponseHandler.java
@@ -2,11 +2,12 @@ package org.stellar.sdk.requests;
 
 import com.google.gson.reflect.TypeToken;
 
-import okhttp3.Response;
-import okhttp3.ResponseBody;
 import org.stellar.sdk.responses.GsonSingleton;
+import org.stellar.sdk.responses.TypedResponse;
 
 import java.io.IOException;
+
+import okhttp3.Response;
 
 public class ResponseHandler<T> {
 
@@ -41,6 +42,9 @@ public class ResponseHandler<T> {
       T object = GsonSingleton.getInstance().fromJson(content, type.getType());
       if (object instanceof org.stellar.sdk.responses.Response) {
         ((org.stellar.sdk.responses.Response) object).setHeaders(response.headers());
+      }
+      if(object instanceof TypedResponse) {
+    	  	((TypedResponse) object).setType(type);
       }
       return object;
     } finally {
--- a/src/main/java/org/stellar/sdk/responses/Page.java
+++ b/src/main/java/org/stellar/sdk/responses/Page.java
@@ -1,27 +1,31 @@
 package org.stellar.sdk.responses;
 
+import static java.util.Objects.requireNonNull;
+
 import com.google.gson.annotations.SerializedName;
 import com.google.gson.reflect.TypeToken;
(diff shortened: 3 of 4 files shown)


nullstyle:
@davejohnclark Thanks for opening this PR!  Could you look into resolving the conflicts?  I'd be happy to merge afterwards.

Also, if you have not yet, please sign our CLA:  https://docs.google.com/forms/d/1g7EF6PERciwn7zfmfke5Sir2n10yddGGSXyZsq98tVY/viewform?usp=send_form

davejohnclark:
Thanks for taking a look.  I've resolved the merge conflicts, but there's an outstanding question on #66  that might be worth resolving before we go any further with this.  I'm happy to make the change here if that's what we want to do.

I do have an observation based on one of the changes that I pulled in.  It's nice to see the http client instance being passed around - big fan of this approach as I mentioned above.  But I'm not sure passing it as an argument to `getNextPage` is necessarily what I'd expect.  That puts the burden on the caller to know how to get the right client instance and pass it in.  I'd prefer to see `Page` encapsulate the http client, as it is really required for the `Page` object to be able to function correctly.  Something similar to the way I'm propagating the type information in the `ResponseHandler` would mean these objects could have the client provided before they are ever returned, insulating the caller from the internals of these objects.

PR #72, opened 2018-01-24
Android support

### LINKED ISSUE #44: The API method for using this SDK in android has an exception
java.lang.NoSuchFieldError: No static field INSTANCE of type Lorg/apache/http/conn/ssl/AllowAllHostnameVerifier; in class Lorg/apache/http/conn/ssl/AllowAllHostnameVerifier; or its superclasses (declaration of 'org.apache.http.conn.ssl.AllowAllHostnameVerifier' appears in /system/framework/framework.jar:classes2.dex)

when I wirte this：“     
    String uri = "https://horizon-testnet.stellar.org";
                    Server server=null;
                    try {
                         server = new Server(uri);
                    }catch (Throwable e){
                        e.printStackTrace();
                        System.out.println("==========异常");
                    }

”then,has an exception,

bartekn:
In PR we remove `org.apache.httpcomponents:httpclient` dependency and start using `com.squareup.okhttp3:okhttp` for all http requests. This allows using this SDK in Android.

Close #44.

android-stellar


eschgi:
Hello,

i have forked some days ago this repository and ported the library to okhttp3.

@bartekn: if you want i can make also a pull request of my implementation or we put the best things of our two implementations together. =)

eschgi:
Hello,

forgot to add the repository link: https://github.com/eschgi/java-stellar-sdk

eschgi:
Hello,

in my implementation i have made some minor api changes, for example i have made some static methods non-static and i've tried to also reduce the use of singletons.

reverbel:
Wow! Thanks for your work, @bartekn.  

I haven't looked at all the changes yet, but what I saw so far looks very good. 

The dependency on Apache HttpClient was the biggest hurdle on the way to supporting Android, and OkHttp is a great choice. However, another (much smaller) showstopper remains: the dependency on commons-codec. Methods such as encodeAsString() (from Base64) do not work on Android.

eschgi:
Hello,

Good work @bartekn! :) I have reviewed your code also and i have some suggestions. Would it be possible that a user of the library can inject his own instance of OkHttpClient and Gson. I think this would be a better design.

Here my implementation of my fork:
* https://github.com/eschgi/java-stellar-sdk/blob/master/src/main/java/org/stellar/sdk/requests/RequestBuilder.java

reverbel:
@bartekn:  A ``Response`` object returned by ``newCall`` contain a ``ResponseBody`` that must be closed. The easiest way to ensure that ``close`` will be invoked on every ``ResponseBody`` is to use "try with resources", as indicated in this page: http://square.github.io/okhttp/3.x/okhttp/okhttp3/ResponseBody.html

Have a look at the convinience method ``executeGetAndHandleResponse``, in file
https://github.com/reverbel/java-stellar-sdk/blob/okhttp/src/main/java/org/stellar/sdk/http/HttpClient.java,
and at method ``submitTransaction``, in file
https://github.com/reverbel/java-stellar-sdk/blob/okhttp/src/main/java/org/stellar/sdk/Server.java.

(Yes, I have also ported java-stellar-sdk to OkHttp...)

reverbel:
@bartekn:  try-with-resources was introduced in Java 1.7. If java-stellar-sdk needs to work with earlier versions of Java, then you will need to use ``try {...} finally {...}`` instead. 

Another thing: file https://github.com/reverbel/java-stellar-sdk/blob/okhttp/src/main/java/org/stellar/sdk/http/HttpClient.java uses a slightly better implementation of the singleton pattern. See  https://stackoverflow.com/questions/70689/what-is-an-efficient-way-to-implement-a-singleton-pattern-in-java.

bartekn:
@eschgi @reverbel wow, guys looks like you've done it before me. 😉  Too bad you didn't submit a PR but, on the bright side, we can pick best parts of each implementation. If you don't mind let's make this PR a main implementation that will be merged eventually.

> However, another (much smaller) showstopper remains: the dependency on commons-codec. 

You're correct, I only tested code related to HTTP requests in Android. Are aware of any other pieces of code that won't work in Android?

Unfortunately, AFAIK there is no way to test library code (like this) in Android environment other that creating a new Android project that utilizes the library. It sucks but it's what we'll have to do eventually.

> Would it be possible that a user of the library can inject his own instance of OkHttpClient and Gson.

Good point, I'll add this.

> A Response object returned by newCall contain a ResponseBody that must be closed. 

You are correct but we don't have to do it because we don't use the stream directly. In all cases we call [`string()`](http://square.github.io/okhttp/3.x/okhttp/okhttp3/ResponseBody.html#string--) method which automatically closes the `ResponseBody`.

> try-with-resources was introduced in Java 1.7. If java-stellar-sdk needs to work with earlier versions of Java, then you will need to use try {...} finally {...} instead.

I think we don't use this statement anywhere in the SDK. Does any of our dependencies use it? If so and there's no alternative that works in both Android and Java application then we'll probably have to drop Java <1.7 support.

> Another thing: file https://github.com/reverbel/java-stellar-sdk/blob/okhttp/src/main/java/org/stellar/sdk/http/HttpClient.java uses a slightly better implementation of the singleton pattern. See https://stackoverflow.com/questions/70689/what-is-an-efficient-way-to-implement-a-singleton-pattern-in-java.

I think double checked locking approach is OK but enum seems to be just simpler. I'll switch to enum, thanks!

reverbel:
> Are aware of any other pieces of code that won't work in Android?

No, I am not. I think we should be fine after dropping httpclient and codec.

> In all cases we call string() method which automatically closes the ResponseBody.

Oh, you are right. I did not noticed that.

> I think we don't use this statement anywhere in the SDK. 

Yes, I know that.

> Does any of our dependencies use it?

I don't think so. Mentioned the Java version >= 1.7 limitation just because the links in my previous comment (a couple of files of my own port of the SDK) use try-with-resources. And now I see that I didn't have to use that statement, as I also called string() in all cases.

eschgi:
> Are aware of any other pieces of code that won't work in Android?

The library uses a older version of the guava library. What i've read that there exists a special version for android that is compatible with java 7. The jre version only supports java 8 but that is a problem for android. Eventually it would be better if the dependency to the guava library could be removed. For more information follow this link: [GitHub](https://github.com/google/guava/blob/master/README.md)

```
dependencies {
  compile 'com.google.guava:guava:23.6-jre'
  // or, for Android:
  compile 'com.google.guava:guava:23.6-android'
}
```

bartekn:
If I'm not missing something, we're currently using the following classes from Guava:

* A few from com.google.gson.*
* com.google.common.base.Preconditions.checkNotNull
* com.google.common.io.BaseEncoding

and all of them work in Android.

eschgi:
Yes, that is right. I've seen also no problems with the guava version 18.0. I've only tried to update guava to the newest version without success.

Another thing:

I would add a class that could be named HttpRequestExecutor that would have following api (see below). This class should be used every time when a http call must be done. I have copied this code partially from my fork:
* https://github.com/eschgi/java-stellar-sdk/blob/master/src/main/java/org/stellar/sdk/requests/RequestBuilder.java

@bartekn: What do you think about this approach?

```
public class HttpRequestExecutor {

  private OkHttpClient httpClient;
  private Gson gson;

  public HttpRequestExecutor(OkHttpClient httpClient, Gson gson) {
    this.httpClient = httpClient;
    this.gson = gson;
  }  

  public <TResponse> TResponse get(final HttpUrl url,
                                      final Class<TResponse> classOfResponse) throws IOException {
    return request(new Request.Builder().url(url).get(), classOfResponse);
  }

  public <TResponse> TResponse get(final HttpUrl url,
                                      final Type typeOfResponse) throws IOException {
    return request(new Request.Builder().url(url).get(), typeOfResponse);
  }

  public <TResponse> TResponse request(final Request.Builder httpRequestBuilder,
                                          final Class<TResponse> classOfResponse) throws IOException {
    return request(httpRequestBuilder, classOfResponse, null);
  }

  public <TResponse> TResponse request(final Request.Builder httpRequestBuilder,
                                          final Type typeOfResponse) throws IOException {
    return request(httpRequestBuilder, null, typeOfResponse);
  }

  private <TResponse> TResponse request(final Request.Builder httpRequestBuilder,
                                        final Class<TResponse> classOfResponse,
                                        final Type typeOfResponse) throws IOException {
    okhttp3.Request httpRequest = httpRequestBuilder.build();
    okhttp3.Response httpResponse = httpClient.newCall(httpRequest).execute();
    if (httpResponse.isSuccessful()) {
      ResponseBody httpResponseBody = httpResponse.body();
      if (httpResponseBody == null) {
        throw new ClientProtocolException("Response contains no content");
      }

      String json = httpResponseBody.string();

      /*
       * "Generics on a type are typically erased at runtime, except when the type is compiled with the
       * generic parameter bound. In that case, the compiler inserts the generic type information into
       * the compiled class. In other cases, that is not possible."
       * More info: http://stackoverflow.com/a/14506181
       */
      TResponse response;
      if (typeOfResponse != null) {
        response = this.gson.fromJson(json, typeOfResponse);
      } else {
        response = this.gson.fromJson(json, classOfResponse);
      }

      if (response instanceof Response) {
        ((Response) response).setHeaders(httpResponse.headers());
      }

      return response;
    } else {
      int statusCode = httpResponse.code();

      // Too Many Requests
      if (statusCode == 429) {
        String retryAfterValue = httpResponse.header("Retry-After");
        if (retryAfterValue != null) {
          int retryAfter = Integer.parseInt(retryAfterValue);
          throw new TooManyRequestsException(retryAfter);
        } else {
          throw new TooManyRequestsException(0);
        }
      }

      String statusMessage = httpResponse.message();
      throw new HttpResponseException(statusCode, statusMessage);
    }
  }
}
```

eschgi:
Hello,

when i've ported the library to okhttp3 i wanted to add also the possibility to support asynchronous http requests. But this is actually difficult without a breaking change.

@bartekn : It would be possible to change the request builder api's as follows: 

```java
// Every request builder returns only Request objects that can be executed by the RequestExecutor
Request<Page<AccountResponse>> request = AccountsRequestBuilder().limit(10).build();

public final class Request<TResponse> {

  private Request.Builder httpRequestBuilder;
  private Class<TResponse> classOfResponse;
  private Type typeOfResponse;
  
  public Request(final Request.Builder httpRequestBuilder,
                           final Class<TResponse> classOfResponse,
                           final Type typeOfResponse) {
    this.httpRequestBuilder = httpRequestBuilder;
    this.classOfResponse = classOfResponse;
    this.typeOfResponse = typeOfResponse;
  }
  
  public Request.Builder getHttpRequestBuilder() {
    return this.httpRequestBuilder;
  }

  public Class<TResponse> getClassOfResponse() {
    return this.classOfResponse;
  }

  public Type getTypeOfResponse() {
    return this.typeOfResponse;
  }
  
}

RequestExecutor requestExecutor = new RequestExecutor(httpClient, gson);

// Execute the request synchronously
Page<AccountResponse> result =  requestExecutor.execute(request);

// Execute the request asynchronously
requestExecutor.executeAsync(request, new ResponseCallback<Page<AccountResponse>>() {
  
  @Override
  public void handleResult(Page<AccountResponse> response) {
  }

  @Override
  public void handleError(Exception e) {
  }

});
```

I could also help to implement this approach.

Kind regards
eschgi

PegasusWallet:
Nice thank you for your PR @bartekn I was going to make one this weekend to enable Android support.
In the mean time  for our App we have stripped httpcomponents:httpclient and replaced by okhttp
We didn't notice any issues SDK wise.

The only issue we had is that proguard was not happy with javax.inject. The code bellow resolves it.

 compile ('org.glassfish.jersey.core:jersey-client:2.22.1'){
        exclude module: 'javax.inject'
    }
    compile ('org.glassfish.jersey.media:jersey-media-sse:2.22.1'){
        exclude module: 'javax.inject'
    }

src/main/java/org/stellar/sdk/Server.java
```diff
@@ -115,41 +109,22 @@ public TransactionsRequestBuilder transactions() {
...
-                IOUtils.copy(responseStream, writer, StandardCharsets.UTF_8);
-                String responseString = writer.toString();
-                SubmitTransactionResponse submitTransactionResponse = GsonSingleton.getInstance().fromJson(responseString, SubmitTransactionResponse.class);
-                return submitTransactionResponse;
-            } finally {
-                responseStream.close();
-            }
+        HttpUrl transactionsURI = serverURI.newBuilder().addPathSegment("transactions").build();
+
+        RequestBody requestBody = new MultipartBody.Builder()
```
PegasusWallet: RequestBody formBody = new FormBody.Builder().add("tx", transaction.toEnvelopeXdrBase64()).build();
Should suffice

bartekn: Better, thanks!

bartekn:
> The only issue we had is that proguard was not happy with javax.inject. The code bellow resolves it.

I resolved it but adding this to `build.gradle`:
```
duplicatesStrategy DuplicatesStrategy.EXCLUDE
```

bartekn:
> when i've ported the library to okhttp3 i wanted to add also the possibility to support asynchronous http requests. But this is actually difficult without a breaking change.

Let's add it via another PR once this one is merged.

src/main/java/org/stellar/sdk/Server.java
```diff
@@ -115,41 +109,18 @@ public TransactionsRequestBuilder transactions() {
...
-                responseStream.close();
-            }
+        HttpUrl transactionsURI = serverURI.newBuilder().addPathSegment("transactions").build();
+        RequestBody requestBody = new FormBody.Builder().add("tx", transaction.toEnvelopeXdrBase64()).build();
+        Request submitTransactionRequest = new Request.Builder().url(transactionsURI).post(requestBody).build();
+        Response response = this.httpClient.newCall(submitTransactionRequest).execute();
+
+        ResponseBody responseBody = response.body();
+
+        if (responseBody == null) {
```
PegasusWallet: In that case the body is never closed and will lead to a memory leak

src/main/java/org/stellar/sdk/federation/FederationServer.java
```diff
@@ -76,39 +74,34 @@ public FederationServer(String serverUri, InternetDomainName domain) {
...
-      uriBuilder.append("/.well-known/stellar.toml");
-      stellarTomlUri = new URI(uriBuilder.toString());
-    } catch (URISyntaxException e) {
-      throw new RuntimeException(e);
+      response = httpClient.newCall(request).execute();
+    } catch (IOException e) {
+      throw new ConnectionErrorException();
+    }
+
+    if (response.code() >= 300) {
```
PegasusWallet: ResponseBody will never be closed -> memory leak

src/main/java/org/stellar/sdk/federation/FederationServer.java
```diff
@@ -76,39 +74,34 @@ public FederationServer(String serverUri, InternetDomainName domain) {
...
-
-                  StringWriter writer = new StringWriter();
-                  InputStream content = entity.getContent();
-                  IOUtils.copy(content, writer);
-                  return new Toml().read(writer.toString());
-                }
-              });
+      stellarToml = new Toml().read(responseBody.string());
     } catch (IOException e) {
       throw new ConnectionErrorException();
```
PegasusWallet: same here. really the best way to handle this is try,catch and finally.

src/main/java/org/stellar/sdk/Server.java
bartekn: You can't close `ResponseBody` which is `null`. And if it's not `null`, it's automatically closed because we're using `string()` method. This has been explained in another comment in this PR.

src/main/java/org/stellar/sdk/federation/FederationServer.java
bartekn: Check my previous comment. `response` is `Response` object, not `ResponseBody` object. And we don't have to close it because we're using `string()` method that closes it.

bartekn: Actually, you may be right. OkHttp docs are not super clear on if `Response` must be closed even if `ResponseBody` was not requested/processed.

bartekn: OK, it looks like we need to close the `Response` even if it's body is not used: https://github.com/square/okhttp/issues/2311. Will fix this, thanks for catching!

CarlosRoldanx:
Hello @bartekn !

Sorry for my low knowledge. On this PR, what I have to do to make it work on Android.

I guess me next steps are downloading these files,  implementing to the Android project, create a jar and then add it as a dependence. 

Am I right?

Thanks.

bartekn:
@CarlosRoldanx yes, you'd have to pull `android` branch from this repo and build a JAR using Graddle (`./graddlew jar`). But we're going to release a new version that will include this PR on Monday so you can wait if you don't want to do it yourself.

reverbel:
Off-topic: Now that we will have the SDK working on Android, it would be nice to be able to extract a Transaction instance from a base64-encoded XDR envelope. The lack of this ability is the remaining showstopper for a very convenient way of withdrawing funds from an off-line cold wallet:
a mobile app that scans a QR code with a transaction envelope and sends the transaction out to horizon.

@bartekn: Have you had a chance to look at PR #69?

bartekn:
@reverbel reviewed your code. Let's move the discussion there.

reverbel:
@bartekn Great, thanks.

eschgi:
Hello,

i've also reviewed the code and this is a little bit inconsistent (see code below).

```java

  public AccountsRequestBuilder(OkHttpClient httpClient, HttpUrl serverURI) {
    super(httpClient, serverURI, "accounts");
  }
  
  /**
   * Requests specific <code>uri</code> and returns {@link Page} of {@link AccountResponse}.
   * This method is helpful for getting the next set of results.
   * @return {@link Page} of {@link AccountResponse}
   * @throws TooManyRequestsException when too many requests were sent to the Horizon server.
   * @throws IOException
   */
  public static Page<AccountResponse> execute(HttpUrl uri) throws IOException, TooManyRequestsException {
    TypeToken type = new TypeToken<Page<AccountResponse>>() {};
    ResponseHandler<Page<AccountResponse>> responseHandler = new ResponseHandler<Page<AccountResponse>>(type);

    OkHttpClient client = new OkHttpClient(); // <-- here is used another instance of OkHttpClient, but it would be better if here is used the instance of the AccountsRequestBuilder class
    Request request = new Request.Builder().get().url(uri).build();
    Response response = client.newCall(request).execute();

    return responseHandler.handleResponse(response);
  }
```

The static method execute creates every time a new instance of the OkHttpClient but it would be better if the instance of the class is used.

In my opinion this should not be so and this behavior is not expected. 

Thanks

eschgi:
Hello,

i've found a compatibility problem with android sdk < 19.

```java
public MemoText(String text) {
    this.text = checkNotNull(text, "text cannot be null");

    int length = text.getBytes((StandardCharsets.UTF_8)).length;
    if (length > 28) {
      throw new MemoTooLongException("text must be <= 28 bytes. length=" + String.valueOf(length));
    }
  }
```

StandardCharsets.UTF_8 is only available on android with sdk version >= 19.

For more informations here a link:
* https://stackoverflow.com/questions/32102166/standardcharsets-utf-8-on-lower-api-lower-than-19

The fix could look like this:

```java
public MemoText(String text) {
    this.text = checkNotNull(text, "text cannot be null");

    int length = text.getBytes("UTF-8").length;
    if (length > 28) {
      throw new MemoTooLongException("text must be <= 28 bytes. length=" + String.valueOf(length));
    }
  }
```

Kind regards
eschgi

bartekn:
> The static method execute creates every time a new instance of the OkHttpClient but it would be better if the instance of the class is used.

@eschgi I think the code you're checking is outdated. Check the latest version.

> StandardCharsets.UTF_8 is only available on android with sdk version >= 19.

Thanks, fixed!

eschgi:
@bartekn: Hello, i reviewed the code again in the android branch. The AccountsRequestBuilder has still a static execute method that uses his own OkHttpClient.

bartekn:
> @bartekn: Hello, i reviewed the code again in the android branch. The AccountsRequestBuilder has still a static execute method that uses his own OkHttpClient.

You're right, my mistake. Fixed!

build.gradle
```diff
@@ -32,11 +33,10 @@ repositories {
 }
 
 dependencies {
-    compile 'commons-codec:commons-codec:1.5'
-    compile 'org.apache.httpcomponents:httpclient:4.5.3'
-    compile 'org.apache.httpcomponents:fluent-hc:4.5.3'
+    compile 'com.squareup.okhttp3:okhttp:3.9.1'
+    compile 'com.squareup.okhttp3:mockwebserver:3.9.1'
```
PegasusWallet: testCompile. Leaving it as compile makes proguard scream and my build fails

bartekn: Fixed, thanks!

CarlosRoldanx:
Hey!

So basically the new version not just is available for Android, It also comes with upgrades. Right?

bartekn:
What do you mean by upgrades?

CarlosRoldanx:
As you commented before the issue related with The AccountsRequestBuilder.

So the new Java SDK version comes with more updates than just the Android support, right?

bartekn:
Yes, you can find complete list of changes in [Releases](https://github.com/stellar/java-stellar-sdk/releases/tag/0.2.0) page or in [CHANGELOG](https://github.com/stellar/java-stellar-sdk/blob/master/CHANGELOG.md) file.

CarlosRoldanx:
Sweet! Thanks @bartekn

PR #97, opened 2018-05-08
Fixes exception thrown from ManageOfferOperation.fromXDR for certain offers

efritze:
The following valid XDR throws an ArithmeticException exception when attempting to decode:

```
@Test
    public void testManageOfferOperationPrice() throws IOException, FormatException {

        String transactionEnvelopeXdr = "AAAAACq1Ixcw1fchtF5aLTSw1zaYAYjb3WbBRd4jqYJKThB9AAAAZAA8tDoAAAAaAAAAAAAAAAAAAAABAAAAAAAAAAMAAAAAAAAAAUJUQwAAAAAAmyMegjdqwy59ijGMyd+sKLgoCfagDexhF17wyd36y2oAAAAABfXhAAABJKNegYQdAAAAAAAAAAAAAAAAAAAAAA==";

        BaseEncoding base64Encoding = BaseEncoding.base64();
        byte[] bytes = base64Encoding.decode(transactionEnvelopeXdr);

        org.stellar.sdk.xdr.TransactionEnvelope transactionEnvelope = TransactionEnvelope.decode(new XdrDataInputStream(new ByteArrayInputStream(bytes)));
        org.stellar.sdk.xdr.Transaction xdrTransaction = transactionEnvelope.getTx();     
        
        ManageOfferOperation manageOfferOperation = (ManageOfferOperation)Operation.fromXdr(xdrTransaction.getOperations()[0]); //Arithmetic exception thrown here.
        
        Price priceExpected = new Price(74915, 1585546269);
        Price priceObtained = Price.fromString(manageOfferOperation.getPrice());
        
        assertEquals(priceExpected, priceObtained);
       
        
    }
```

This is principally because https://github.com/stellar/java-stellar-sdk/blob/master/src/main/java/org/stellar/sdk/ManageOfferOperation.java#L115 throws the following exception:

**ArithmeticException: "Non-terminating decimal expansion; no exact representable decimal result" **

This pull request fixes the issue, and also changes the type of the price field member in  ManageOfferOperation from string to Price. This is necessary for projects that demand a more accurate representation of an offer's price.

### DIFF
--- a/src/main/java/org/stellar/sdk/ManageOfferOperation.java
+++ b/src/main/java/org/stellar/sdk/ManageOfferOperation.java
@@ -9,6 +9,7 @@ import org.stellar.sdk.xdr.Uint64;
 import java.math.BigDecimal;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import java.math.RoundingMode;
 
 /**
  * Represents <a href="https://www.stellar.org/developers/learn/concepts/list-of-operations.html#manage-offer" target="_blank">ManageOffer</a> operation.
@@ -19,10 +20,10 @@ public class ManageOfferOperation extends Operation {
   private final Asset selling;
   private final Asset buying;
   private final String amount;
-  private final String price;
+  private final Price price;
   private final long offerId;
 
-  private ManageOfferOperation(Asset selling, Asset buying, String amount, String price, long offerId) {
+  private ManageOfferOperation(Asset selling, Asset buying, String amount, Price price, long offerId) {
     this.selling = checkNotNull(selling, "selling cannot be null");
     this.buying = checkNotNull(buying, "buying cannot be null");
     this.amount = checkNotNull(amount, "amount cannot be null");
@@ -55,7 +56,12 @@ public class ManageOfferOperation extends Operation {
   /**
    * Price of 1 unit of selling in terms of buying.
    */
+  @Deprecated
   public String getPrice() {
+    return new BigDecimal(price.getNumerator()).divide(new BigDecimal(price.getDenominator()),8, RoundingMode.DOWN).toPlainString();
+  }
+  
+  public Price getPriceR() {
     return price;
   }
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/stellar/sdk/ManageOfferOperation.java
```diff
@@ -19,10 +20,10 @@
   private final Asset selling;
   private final Asset buying;
   private final String amount;
-  private final String price;
+  private final Price price;
   private final long offerId;
 
-  private ManageOfferOperation(Asset selling, Asset buying, String amount, String price, long offerId) {
+  private ManageOfferOperation(Asset selling, Asset buying, String amount, Price price, long offerId) {
```
bartekn: Can we overload the method (leave the existing methods and add a new methods with `Price`) for backward compatibility?

src/main/java/org/stellar/sdk/ManageOfferOperation.java
```diff
@@ -55,7 +56,12 @@ public String getAmount() {
   /**
    * Price of 1 unit of selling in terms of buying.
    */
+  @Deprecated
   public String getPrice() {
+    return new BigDecimal(price.getNumerator()).divide(new BigDecimal(price.getDenominator()),8, RoundingMode.DOWN).toPlainString();
+  }
+  
+  public Price getPriceR() {
```
bartekn: Can we leave `public String getPrice` as is for backward compatibility?

src/main/java/org/stellar/sdk/ManageOfferOperation.java
```diff
@@ -125,10 +130,28 @@ public long getOfferId() {
...
+    /**
+     * Creates a new ManageOffer builder. If you want to update existing offer use
+     * {@link org.stellar.sdk.ManageOfferOperation.Builder#setOfferId(long)}.
+     * @param selling The asset being sold in this operation
+     * @param buying The asset being bought in this operation
+     * @param amount Amount of selling being sold.
+     * @param price Price of 1 unit of selling in terms of buying.
+     * @throws ArithmeticException when amount has more than 7 decimal places.
+     */
+    @Deprecated
```
bartekn: I think we shouldn't deprecate this method. It's OK to use `String` price in many cases.

efritze:
Moved to https://github.com/stellar/java-stellar-sdk/pull/293

PR #105, opened 2018-06-07
TradeResponse add price attribute and set and get method

hubs:


bartekn:
Closing in favour of #106.

PR #108, opened 2018-06-19
use org.stellar.sdk.Price instead of String in org.stellar.sdk.ManageOfferOperation

mayan31370:
use org.stellar.sdk.Price instead of String in org.stellar.sdk.ManageOfferOperation

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,7 +2,7 @@ apply plugin: 'java'
 apply plugin: 'ivy-publish'
 
 sourceCompatibility = 1.6
-version = '0.2.0'
+version = '0.2.1'
 group = 'stellar'
 
 jar {


src/main/java/org/stellar/sdk/ManageOfferOperation.java
```diff
@@ -129,12 +132,28 @@ public Builder(Asset selling, Asset buying, String amount, String price) {
...
+    public Builder(Asset selling, Asset buying, String amount, Price price) {
+      this.selling = checkNotNull(selling, "selling cannot be null");
+      this.buying = checkNotNull(buying, "buying cannot be null");
+      this.amount = checkNotNull(amount, "amount cannot be null");
+      this.price = price;
     }
 
     /**
      * Sets offer ID. <code>0</code> creates a new offer. Set to existing offer ID to change it.
-     * @param offerId
```
bartekn: Can you revert this?

src/main/java/org/stellar/sdk/ManageOfferOperation.java
```diff
@@ -55,7 +56,7 @@ public String getAmount() {
   /**
    * Price of 1 unit of selling in terms of buying.
    */
-  public String getPrice() {
+  public Price getPrice() {
```
bartekn: Can we make this a non-breaking change? The new method can be: `public Price getPriceObject()` (let me know if you have a better name idea).

src/main/java/org/stellar/sdk/ManageOfferOperation.java
```diff
@@ -129,12 +132,28 @@ public Builder(Asset selling, Asset buying, String amount, String price) {
...
+     * @param buying The asset being bought in this operation
+     * @param amount Amount of selling being sold.
+     * @param price {@link org.stellar.sdk.Price} Price of 1 unit of selling in terms of buying.
+     * @throws ArithmeticException when amount has more than 7 decimal places.
+     */
+    public Builder(Asset selling, Asset buying, String amount, Price price) {
+      this.selling = checkNotNull(selling, "selling cannot be null");
+      this.buying = checkNotNull(buying, "buying cannot be null");
+      this.amount = checkNotNull(amount, "amount cannot be null");
+      this.price = price;
```
bartekn: Would be great to have `checkNotNull` here.

PR #114, opened 2018-07-17
Construct Transaction from xdr.TransactionEnvelope

bartekn:
Currently it's not possible to add a new signature to the existing envelope using Java SDK. Added two methods:
* `Transaction.fromEnvelopeXdr(String envelope)`
* `Transaction.fromEnvelopeXdr(TransactionEnvelope envelope)`

that allow creating `Transaction` object from existing envelope to append a new signatures later.

Added helper functions i.e. `fromXdr` and `equals` to some classes to simplify tests.

### DIFF
--- a/src/test/java/org/stellar/sdk/TransactionTest.java
+++ b/src/test/java/org/stellar/sdk/TransactionTest.java
@@ -41,6 +41,18 @@ public class TransactionTest {
     assertEquals(transaction.getSourceAccount(), source);
     assertEquals(transaction.getSequenceNumber(), sequenceNumber+1);
     assertEquals(transaction.getFee(), 100);
+
+    Transaction transaction2 = Transaction.fromEnvelopeXdr(transaction.toEnvelopeXdr());
+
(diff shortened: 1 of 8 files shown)


src/test/java/org/stellar/sdk/TransactionTest.java
```diff
@@ -41,6 +41,18 @@ public void testBuilderSuccessTestnet() throws FormatException {
     assertEquals(transaction.getSourceAccount(), source);
     assertEquals(transaction.getSequenceNumber(), sequenceNumber+1);
     assertEquals(transaction.getFee(), 100);
+
+    Transaction transaction2 = Transaction.fromEnvelopeXdr(transaction.toEnvelopeXdr());
```
michaelran16: 👍

Kirbyrawr:
Hi, i'm not a contributor to the Java SDK however we were also missing this feature in the C# SDK, i basically ported this pull request and the tests are working ok.

Just in case you want another review.

PR #118, opened 2018-07-23
Fix creating Memo.ReturnHash

### LINKED ISSUE #116: MemoReturnHash#toXdr sets Hash, not RetHash
https://github.com/stellar/java-stellar-sdk/blob/master/src/main/java/org/stellar/sdk/MemoReturnHash.java#L26

bartekn:
Fix #116

### DIFF
--- a/src/test/java/org/stellar/sdk/MemoTest.java
+++ b/src/test/java/org/stellar/sdk/MemoTest.java
@@ -1,13 +1,12 @@
 package org.stellar.sdk;
 
+import com.google.common.io.BaseEncoding;
 import org.junit.Test;
 import org.stellar.sdk.xdr.MemoType;
 
 import java.util.Arrays;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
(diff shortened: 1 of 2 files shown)


src/test/java/org/stellar/sdk/MemoTest.java
```diff
@@ -112,7 +111,10 @@ public void testMemoHashInvalidHex() {
     @Test
     public void testMemoReturnHashSuccess() {
         MemoReturnHash memo = Memo.returnHash("4142434445464748494a4b4c");
-        assertEquals(MemoType.MEMO_RETURN, memo.toXdr().getDiscriminant());
+        org.stellar.sdk.xdr.Memo memoXdr = memo.toXdr();
```
nikhilsaraf: why not import `org.stellar.sdk.xdr.Memo` as a type directly (instead of referencing the full name here)?

bartekn: There's a class with the same name (`Memo`) in `org.stellar.sdk` package where this test file live. It would be a name conflict.

nikhilsaraf: 👍

Issue #122, opened 2018-08-09
classpath issues with Spring Boot 2.x


jillesvangurp:
I've been trying to use the api client in a Spring Boot project. One of the transitive dependencies is pulling in some old version of javax.validator, which is causing a lot of problems with spring boot auto configuration. So I looked at the transitive dependencies and it seems to pull in a lot of stuff, most of which is pretty out of date. 

```
com.github.Stellar:java-stellar-sdk:0.3.0

     com.squareup.okhttp3:okhttp:3.9.1
     com.moandjiezana.toml:toml4j:0.5.1
         com.google.code.gson:gson:2.3.1 ➡ 2.8.5
     org.mockito:mockito-core:2.13.0 ➡ 2.15.0
         net.bytebuddy:byte-buddy:1.7.9 ➡ 1.7.11
         net.bytebuddy:byte-buddy-agent:1.7.9 ➡ 1.7.11
         org.objenesis:objenesis:2.6
     com.google.guava:guava:18.0
     com.google.code.gson:gson:2.4 ➡ 2.8.5
     commons-io:commons-io:2.4
     org.glassfish.jersey.core:jersey-client:2.22.1 ➡ 2.26
         javax.ws.rs:javax.ws.rs-api:2.1
         org.glassfish.jersey.core:jersey-common:2.26
             javax.ws.rs:javax.ws.rs-api:2.1
             javax.annotation:javax.annotation-api:1.2 ➡ 1.3.2
             org.glassfish.hk2.external:javax.inject:2.5.0-b42
             org.glassfish.hk2:osgi-resource-locator:1.0.1
         org.glassfish.hk2.external:javax.inject:2.5.0-b42
     org.glassfish.jersey.media:jersey-media-sse:2.22.1
         org.glassfish.hkject:2.4.0-b31 ➡ 2.5.0-b42
         org.glassfish.jersey.core:jersey-server:2.22.1 ➡ 2.26
             org.glassfish.jersey.core:jersey-common:2.26
             org.glassfish.jersey.core:jersey-client:2.26
             javax.ws.rs:javax.ws.rs-api:2.1
             org.glassfish.jersey.media:jersey-media-jaxb:2.26
                 org.glassfish.jersey.core:jersey-common:2.26
                 org.glassfish.hk2.external:javax.inject:2.5.0-b42
                 org.glassfish.hk2:osgi-resource-locator:1.0.1
             javax.annotation:javax.annotation-api:1.2 ➡ 1.3.2
[...]

PR #119, opened 2018-07-23
Upgrade Gradle

bartekn:
Fix #94.

PR #123, opened 2018-08-09
shade all dependencies to avoid conflicts

jillesvangurp:
Addresses #122

We cannot use the original jar because of Spring Boot autoconfiguring included things, which then fails because they are old versions. Gson and javax.validation are the two problematic dependencies. 

There are several other things in the dependencies that I'd prefer to not have there like jersey and mockito (why is this a compile dependency).

Shading takes care of removing the conflicts for us; so I simply shaded everything. 

I've signed the contributor agreement.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,9 +53,11 @@ dependencies {
 }
 
 jar.enabled = false
+
 assemble.dependsOn(shadowJar);
 
 shadowJar {
+   archiveName 'stellar-sdk.jar'
    relocate 'com','shadow.com'
    relocate 'net','shadow.net'
    relocate 'org.glassfish','shadow.org.glassfish'
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,9 @@
+
+plugins {
+  id "com.github.johnrengelman.shadow" version "2.0.4"
+}
+
+apply plugin: "com.github.johnrengelman.shadow"
 apply plugin: 'java'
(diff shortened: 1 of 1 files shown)


jillesvangurp:
Travis currently breaks because of an out dated gradle. This will require merging of #119 to upgrade gradle. I used gradle 4.9 and it builds fine with that.

bartekn:
#119 is merged. Can you rebase `master`?

jillesvangurp:
@bartekn just updated my branch

cyberluke:
How does the shading work for you? I download current SDK 0.3.3 and have this exact same issue again: https://github.com/stellar/java-stellar-sdk/issues/122

jillesvangurp:
@cyberluke We merged but never released a fix for this.  

We have another fix in the works that removes jersey entirely. @bartekn we could actually get rid of this shading if we merge that ( #147 )

@cyberluke, give my fork a spin  by depending on: 'com.github.Inbot:java-stellar-sdk:v0.3.2.1-inbot-get-rid-of-jersey'

cyberluke:
@jillesvangurp I just forked and sent pull request that solves all my issues. The main fuck up was because of using guava android, not guava jre: https://github.com/stellar/java-stellar-sdk/pull/152

jillesvangurp:
Try my fork, it shades properly. I use this with spring boot 2.x.

Issue #129, opened 2018-08-22
Issue fetching transactions for ledger


jonlorusso:
Up to and including version `0.3.2`, the following `MemoId` causes a `NumberFormatException` (greater than `Long.MAX_VALUE`)

```
java.lang.NumberFormatException: For input string: "10048071741004807174"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Long.parseLong(Long.java:592)
	at java.lang.Long.parseLong(Long.java:631)
	at org.stellar.sdk.responses.TransactionDeserializer.deserialize(TransactionDeserializer.java:46)
	at org.stellar.sdk.responses.TransactionDeserializer.deserialize(TransactionDeserializer.java:16)
	at com.google.gson.TreeTypeAdapter.read(TreeTypeAdapter.java:58)
	at com.google.gson.internal.bind.TypeAdapterRuntimeTypeWrapper.read(TypeAdapterRuntimeTypeWrapper.java:40)
	at com.google.gson.internal.bind.CollectionTypeAdapterFactory$Adapter.read(CollectionTypeAdapterFactory.java:81)
	at com.google.gson.internal.bind.CollectionTypeAdapterFactory$Adapter.read(CollectionTypeAdapterFactory.java:60)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.read(ReflectiveTypeAdapterFactory.java:117)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:217)
	at com.google.gson.Gson.fromJson(Gson.java:814)
	at com.google.gson.Gson.fromJson(Gson.java:879)
	at org.stellar.sdk.responses.PageDeserializer.deserialize(PageDeserializer.java:49)
	at org.stellar.sdk.responses.PageDeserializer.deserialize(PageDeserializer.java:19)
	at com.google.gson.TreeTypeAdapter.read(TreeTypeAdapter.java:58)
	at com.google.gson.Gson.fromJson(Gson.java:814)
	at com.google.gson.Gson.fromJson(Gson.java:779)
	at com.google.gson.Gson.fromJson(Gson.java:728)
	at org.stellar.sdk.requests.ResponseHandler.handleResponse(ResponseHandler.java:42)
	at org.stellar.sdk.requests.TransactionsRequestBuilder.execute(TransactionsRequestBuilder.java:94)
[...]

PR #130, opened 2018-08-22
Treat memo id as an unsigned integer

jonlorusso:
Fixes https://github.com/stellar/java-stellar-sdk/issues/129

### DIFF
--- a/src/main/java/org/stellar/sdk/MemoId.java
+++ b/src/main/java/org/stellar/sdk/MemoId.java
@@ -10,6 +10,9 @@ public class MemoId extends Memo {
   private long id;
 
   public MemoId(long id) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/stellar/sdk/MemoId.java
```diff
@@ -10,9 +10,6 @@
   private long id;
 
   public MemoId(long id) {
-    if (id < 0) {
-      throw new IllegalArgumentException("id must be a positive number");
-    }
```
bartekn: Please revert this change. Memo [must](https://github.com/stellar/stellar-core/blob/master/src/xdr/Stellar-transaction.x#L291) be a positive number (or zero).

PR #132, opened 2018-08-28
Fix shading and upgrade dependencies

jillesvangurp:
The shading pr merged earlier still included all dependencies in the pom.xml, which defeats the purpose.

- I fixed this with some gradle config to customize the pom file to not do that. 
- I also updated all the dependencies to their latest version
- I removed ivy-publish in favor of the maven plugin (supported  by the shadow plugin) and added two plugins to make examining the dependencies easier.
- I moved mockito to the testCompile classpath. I assume this was there by mistake.

To be sure this now works correctly, I've released my fork on jitpack:

```
    implementation 'com.github.Inbot:java-stellar-sdk:v0.3.2.1-inbot-shaded-7'
```

@bartekn I'm depending on this in my kotlin wrapper for the SDK; as soon as you merge and cut a new release I can  switch to that.

To test this locally in the future, run

```
./gradlew clean -xtest install
```

Then check the contents of `~/.m2/repository/stellar/java-stellar-sdk/0.3.2`

It should list something like:

```
-rw-r--r--  1 jillesvangurp  staff   6.5M Aug 28 11:52 java-stellar-sdk-0.3.2-all.jar
-rw-r--r--  1 jillesvangurp  staff   6.5M Aug 28 11:52 java-stellar-sdk-0.3.2.jar
-rw-r--r--  1 jillesvangurp  staff   396B Aug 28 11:52 java-stellar-sdk-0.3.2.pom
```

The important bit here is that we have a fat jar that has relocated dependencies and that the pom file does not have any dependencies (included in the jar). I also examine the contents of the jar to determine which packages need shading. So, when adding dependencies in the future, be sure to check they are relocated.


jillesvangurp:
@bartekn I just merged latest changes from master. Would be nice to get this in now.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Fri Sep 09 23:28:22 CEST 2016
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.9-bin.zip
```
bartekn: I downloaded this file and noticed that `gradle/wrapper/gradle-wrapper.jar` files differ. Why? Other than that LGTM!

jillesvangurp:
I upgraded gradle. 4.10 should be out by now so we could upgrade it again.

Jilles

bartekn:
I understand that it was because of the Gradle upgrade but the file I downloaded (the same version) is binary different that the one in this PR.

jillesvangurp:
@bartekn I've merged all the latest changes on master; minor conflict on build.gradle. I have no explanation for the binary differences on the jar. But to be sure, just updated once more to gradle 4.10.2

jillesvangurp:
Using homebrew installed gradle and using `gradle wrapper`

PR #147, opened 2018-10-26
use okhttp-sse and drop jersey dependency

### LINKED ISSUE #141: streaming apis are currently broken
I did some digging yesterday to figure out why the streaming APIs are not working currently in with the java sdk.

I discovered with a debugger that jersey is throwing an IllegalStateException stating that InjectionManagerFactory cannot be found. This gets caught inside jersey and results in the connection being closed right away after opening it. Events never make it to the event listener

This may be related to the shaded and relocated jar that was introduced. The class exists but has a different package name because of this. Somehow it does not pick it up and it fails without any log messages or visible exceptions. . 

In general, Jersey is doing all sorts of nasty things with reflection, bundling its own dependency injection framework, and pulling in all sorts of crazy  dependencies that are guaranteed to be a headache with e.g. Spring Boot.  I also think this is a bad idea on Android. This is the reason why I did #123; this avoids these conflicts.

A good long term solution would be to get rid of Jersey and instead built a simple solution around okhttp with minimal dependencies. There's absolutely no need to use jersey and IMHO using it should be avoided for client usage. I think it's primary use case is integration tests. 

I imagine this would be doable with in a day or so. Most of the API seems to be doing simple GETs and picking apart the responses with gson (just like is done right now). I think the only POST api is probably the submit transaction one. That too should be straightforward to do. You can do that in couple of lines of code with httpclient. 

Streaming is a bit more work, mainly because you need a background thread that keeps on reading the response and reconnects when needed. But that too is simple using an executor and the existing callback interface. This is basically what Jersey does as well but in a very convoluted way. Aside from that, all you need to do is add a single http header to trigger SSE on the server and a simple parser that picks apart the lines with the events.

There are a couple of short term workarounds. 1) do polling with the non streaming API. 2) reimplement just the streaming endpoints around okhttp as described above. Basically, should not be that hard to do as SSE is a very straightforward protocol. 3) roll back the shading for now and break compatibility with Spring boot 2.x (cannot have jersey on the classpath because of conflicting dependencies).

I'm willing to help on this of course but given how busy you guys are currently and the slow progress on my other PRs I would like to discuss upfront whether this is desirable/welcome now. I'm happy to continue using my fork for now. I implemented a workaround in my kotlin library based on polling that works well enough (Kotlin sequences and coroutines are very nice for this).

jillesvangurp:
Fixes #141 

- drops jersey support
- adds okhttp-sse instead
- reimplements streaming using that
- new streaming methods return the okhttp-sse of EventSource


jillesvangurp:
This is still WIP, tests are missing (there was no test coverage for the old methods either). So, please help test and review. I'll do that as well. Meanwhile, please review the changes. 

Another issue is that  some of the implementations I adapted looked like they were copy pasted and possibly wrong. E.g. PaymentsRequestBuilder was streaming OperationsResponse instead of PaymentOperationsResponse. I've not changed that and left it as it was.

@bartekn it may be a good idea to use docker-compose to launch a standalone chain and run some integration tests against that. I've done this for my kotlin wrapper.

See here for an example:
https://github.com/Inbot/inbot-stellar-kotlin-wrapper/blob/master/build.gradle#L105

This would require enabling docker support on travis. It's been a while since I did that but should be possible as well. 

Alternatively, we can write a simple test and point it at horizon testnet. Should be OK as long as it doesn't run too often.

jillesvangurp:
Sadly does not yet work. Horizon is returning a mimetype of (application/hal+json; charset=utf-8) that okhttp-sse does not like. It seems to check for text/eventstream

This API is marked as experimental on the okhttp side. I'll see if I can workaround this in some way.

bartekn:
> Sadly does not yet work. Horizon is returning a mimetype of (application/hal+json; charset=utf-8) that okhttp-sse does not like. It seems to check for text/eventstream

It returns the `Content-Type: text/evenstream` header, you can easily check it like:
```
curl -v -H "Accept: text/event-stream" "https://horizon-testnet.stellar.org/ledgers?cursor=now"
```
However, Horizon requires `Accept: text/event-stream` request header to start a streaming connection.

jillesvangurp:
@bartekn I have a working version now. Turns out you have to manually set the header on the request with okhttp-sse.

A few changes:

- I've  implemented an `SSEManager` class that orchestrates the reconnecting and stopping logic. 
- There's a very simplistic test that verifies stuff streams from horizon. It's ignored by default to prevent needless API calls/running into rate limits: `org.stellar.sdk.StreamingSmokeTest#shouldStreamPaymentsFromTestNet`

From my point of view this seems kind of a convoluted pattern; but it works. The alternative of simply paging normal API requests and keeping track of the cursor would be about as complicated. This is actually what I implemented in out kotlin wrapper. Very straightforward to do with kotlin sequences.

Another issue is that running background threads pretty much is a bad idea on Android.

Let me know if you like this approach or if you would like changes to this.

src/main/java/org/stellar/sdk/requests/SSEUtils.java
```diff
@@ -0,0 +1,72 @@
...
+import okhttp3.internal.sse.RealEventSource;
+import okhttp3.sse.EventSource;
+import okhttp3.sse.EventSourceListener;
+import org.stellar.sdk.responses.GsonSingleton;
+
+import javax.annotation.Nullable;
+import java.net.SocketException;
+
+
+public class SSEUtils {
```
bartekn: This probably shouldn't be `public`. Also, maybe instead of this class we could have a static method (`create`?) that creates a new stream on `SSEManager`?

src/main/java/org/stellar/sdk/requests/SSEUtils.java
```diff
@@ -0,0 +1,72 @@
...
+  public static <T extends org.stellar.sdk.responses.Response> SSEManager<T> stream(final OkHttpClient okHttpClient, final RequestBuilder requestBuilder, final Class<T> clazz, final EventListener<T> listener) {
```
bartekn: Please rename `clazz` to `responseClass` everywhere.

src/main/java/org/stellar/sdk/requests/SSEManager.java
```diff
@@ -0,0 +1,81 @@
...
+import java.io.Closeable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+interface CloseListener {
+  public void closed(EventSource source);
+}
+
+public class SSEManager<T extends org.stellar.sdk.responses.Response> implements Closeable {
```
bartekn: Let me know what do you think about this:

This class represents a single stream so probably `SSEStream` is a better name. When it comes to a single stream, I think it should only have a single method: `close()` that developer can call when stream is no longer needed. The stream should make use of `id` field in a stream and use this in `Last-Event-ID` request header when making a new connection after Horizon closes the connection. If the response code is not in `2xx` range or stream sends `event: err` event, it should stop and call `onError` listener.

This is defined in [SSE spec](https://www.w3.org/TR/2009/WD-eventsource-20090421/).

src/main/java/org/stellar/sdk/requests/SSEUtils.java
```diff
@@ -0,0 +1,72 @@
...
+            throw new IllegalStateException("Failed " + code, t);
+          }
+        } else {
+          throw new IllegalStateException("Failed " + code);
+        }
+      }
+
+      @Override
+      public void onEvent(EventSource eventSource, @Nullable String id, @Nullable String type, String data) {
+        if (data.equals("\"hello\"")) {
```
bartekn: Horizon sends 4 types of messages:
* `event:open`
* `event:close`
* `event:err`

`event` field is empty when sending data. So you should ignore all `open` and `close` events.

src/main/java/org/stellar/sdk/requests/SSEUtils.java
jillesvangurp: I merged this class into the new SSEStream class and renamed the method.

src/main/java/org/stellar/sdk/requests/SSEUtils.java
jillesvangurp: this should be handled by okhttp-sse EventSourceListener. If I got this wrong, we should be seeing parse errors. I think jersey did something similar under the hood. This hello messages is always the first thing on the wire. After that you get proper json.

src/main/java/org/stellar/sdk/requests/SSEManager.java
jillesvangurp: I renamed the class. okhttp-sse follows the spec and has appropriate listener methods. I've extracted the anonymous class for this to a private static inner class so it is a bit easier to follow. Since we use okhttp-sse, I don't see how we can easily get to the last-event-id unless it already does this for us somehow.

But we are tracking the pagingToken that comes back with each event. I've added that to SSEStream with an atomic reference so you can find out which pagingToken was last seen.

src/main/java/org/stellar/sdk/requests/SSEUtils.java
jillesvangurp: done

src/main/java/org/stellar/sdk/requests/SSEManager.java
jillesvangurp: correction, just noticed that the listener actually passes the id. I now track that and set the header.

jillesvangurp:
@bartekn Thanks for your review. I updated the PR and refactored a bit. There is now just one SSEStream class and most methods on it are private.

src/main/java/org/stellar/sdk/requests/SSEUtils.java
bartekn: Horizon will also send [`goodbye`](https://github.com/stellar/go/blob/b3cacf3189c40fc2ff68bd45fadd5716b9ccc83f/services/horizon/internal/render/sse/main.go#L89) event before closing a stream so this must be handled.

src/main/java/org/stellar/sdk/requests/SSEStream.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+    executorService.submit(new Runnable() {
+      @Override
+      public void run() {
+        while (!isStopped.get()) {
+          try {
+            Thread.sleep(200);
+            if (serverSideClosed.get() && !isStopped.get()) {
+              serverSideClosed.set(false); // don't restart until true again
+              restart();
```
bartekn: I think that the following sequence is possible:
1. `serverSideClosed.set(false);`
2. Someone calls `SSEStream.close()`
3. `restart();`

It will leave the stream running in the background. I think synchronization will be much easier to read and understand in this class with `synchronized` block. If I'm not missing anything you need to ensure that:
* Closing the stream,
* Restarting the stream,
* Setting `lastEventId`

are atomic. But to be honest, I'll be totally OK if this class is not thread safe (in outside use) if we add a proper docs. In most use-cases there's a single thread that controls the stream (can close the stream).

src/main/java/org/stellar/sdk/requests/SSEStream.java
```diff
@@ -0,0 +1,180 @@
...
+public class SSEStream<T extends org.stellar.sdk.responses.Response> implements Closeable {
+  private final OkHttpClient okHttpClient;
+  private final RequestBuilder requestBuilder;
+  private final Class<T> responseClass;
+  private final EventListener<T> listener;
+
+  private final AtomicBoolean isStopped = new AtomicBoolean(false);
+  private final AtomicBoolean serverSideClosed = new AtomicBoolean(true); // make sure we start correctly
+  private final AtomicReference<String> pagingToken = new AtomicReference<String>("0");
+  private final AtomicReference<String> lastEventId = new AtomicReference<String>(null);
```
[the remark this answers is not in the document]
jillesvangurp: That shouldn't happen but since we're using threads, I'd use it on principle here just in case. This shortcuts the potential for a lot of subtle concurrency bugs.

src/main/java/org/stellar/sdk/requests/SSEStream.java
```diff
@@ -0,0 +1,180 @@
...
+
+public class SSEStream<T extends org.stellar.sdk.responses.Response> implements Closeable {
+  private final OkHttpClient okHttpClient;
+  private final RequestBuilder requestBuilder;
+  private final Class<T> responseClass;
+  private final EventListener<T> listener;
+
+  private final AtomicBoolean isStopped = new AtomicBoolean(false);
+  private final AtomicBoolean serverSideClosed = new AtomicBoolean(true); // make sure we start correctly
+  private final AtomicReference<String> pagingToken = new AtomicReference<String>("0");
```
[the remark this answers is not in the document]
jillesvangurp: removed it

src/main/java/org/stellar/sdk/requests/SSEStream.java
jillesvangurp: I've added a lock and double check the stopped state now; shouldn't hurt anyone. Together with the atomic state changes, that should take care of a lot of stuff that will indeed probably never happen. But I've done this stuff long enough to err on the safe side here.

Kind of nasty that we need threads here at all. Probably a bad idea on Android. In kotlin or similar languages, you'd probably use a sequence or co-routines to avoid threading. 

A pure Kotlin client might actually be a good idea since it could address native android/ios, web (via wasm), android, and Java. Just a thought.

jillesvangurp:
@bartekn addressed your comments, have another look please.

src/main/java/org/stellar/sdk/requests/SSEUtils.java
jillesvangurp: this is handled now

cyberluke:
@bartekn It was 21 days. Can you review please? You get what you want after one day, but then nothing happens. 8 days ago I post about streaming issue, which is not working in your official sdk on StackExchange: https://stellar.stackexchange.com/questions/1945/subscribing-to-incoming-transaction-does-not-work ...now I tried this and this solves the streaming issue, it should be in main already. The person sent this pull request is highly skilled, he deserves more support.

bartekn:
Many thanks for this @jillesvangurp and sorry for a delay once again!

PR #150, opened 2018-11-08
Add Transaction.Builder.setTimeout method

bartekn:
This PR adds `Transaction.Builder.setTimeout` method to help prevent situations described in https://github.com/stellar/stellar-core/issues/1811. `Transaction.Builder.setTimeout` sets `timebounds.max_time` on a transaction if it is not set yet. It is required to call:
```java
TransactionBuilder.setTimeout(Transaction.Builder.TIMEOUT_INFINITE);
```
if no `timebounds.max_time` should be set to ensure that developers know about possible implications.

### DIFF
--- a/src/main/java/org/stellar/sdk/Transaction.java
+++ b/src/main/java/org/stellar/sdk/Transaction.java
@@ -261,6 +261,9 @@ public class Transaction {
     private Memo mMemo;
     private TimeBounds mTimeBounds;
     List<Operation> mOperations;
+    private boolean timeoutSet;
+
+    public static final long TIMEOUT_INFINITE = 0;
 
     /**
      * Construct a new transaction builder.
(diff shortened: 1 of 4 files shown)


src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -320,10 +323,49 @@ public Builder addTimeBounds(TimeBounds timeBounds) {
...
+     * If you want to be sure to receive the status of the transaction within a given period you should set the
+     * {@link TimeBounds} with <code>maxTime</code> on the transaction (this is what <code>setTimeout</code> does
+     * internally; if there's <code>minTime</code> set but no <code>maxTime</code> it will be added).
+     * Call to <code>Builder.setTimeout</code> is required if Transaction does not have <code>max_time</code> set.
+     * If you don't want to set timeout, use <code>TIMEOUT_INFINITE</code>. In general you should set
+     * <code>TIMEOUT_INFINITE</code> only in smart contracts.
+     * Please note that Horizon may still return <code>504 Gateway Timeout</code> error, even for short timeouts.
+     * In such case you need to resubmit the same transaction again without making any changes to receive a status.
+     * This method is using the machine system time (UTC), make sure it is set correctly.
+     * @param timeout Timeout in seconds.
```
poliha: Hey @bartekn, what happens when `timeout` is negative? Probably default to `TIMEOUT_INFINITE`?

bartekn: Good catch, fixed!

PR #152, opened 2018-11-12
Drop Guava Android workaround, fix Spring Boot 2, fix Swagger 2, add …

cyberluke:
…Java 8 support

*** Advice for build system, dependencies (Spring Boot 2.x, Swagger 2, perhaps any 2018 Java app) ***
Compiling SDK:
```
> Task :compileJava
warning: [options] source value 6 is obsolete and will be removed in a future release
warning: [options] target value 1.6 is obsolete and will be removed in a future release
```

Using SDK with Swagger 2:
```
Caused by: java.lang.NoSuchMethodError: com.google.common.collect.FluentIterable.concat(Ljava/lang/Iterable;Ljava/lang/Iterable;)Lcom/google/common/collect/FluentIterable;
~[java-stellar-sdk-0.3.3.jar:0.3.2]
```

Using SDK with Spring Boot 2:
```
javax.validation api clash on classpath v1 vs v2
```

Really? This is how you develop Java on computer? This is the nightmare and main cause of trouble and shading requirements (which does not work for me, that's why I post here):
```
    // use the android version because we don't want java 8 stuff
    compile 'com.google.guava:guava:26.0-android'
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,7 +14,7 @@ apply plugin: 'maven' // needed to add the install task for jitpack.io
 apply plugin: 'com.github.ben-manes.versions' // gradle dependencyUpdates -Drevision=release
 apply plugin: 'project-report' // gradle htmlDependencyReport
 
-sourceCompatibility = 1.6
+sourceCompatibility = 1.8
 version = '0.3.2'
 group = 'stellar'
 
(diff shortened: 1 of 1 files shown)


jillesvangurp:
This breaks compatibility with Android. 

There are some known issues with the shading that were addressed in #132. I don't think a new release was tagged with that so the current release definitely does not shade correctly. 

#147 gets rid of jersey and a lot of dependencies. It still depends on guava-android though. 

We could consider replacing that with something else. The main thing would be finding a replacement for base64 that does not require an android specific library. I believe commons-codec is also java-8. 
Simply copying the some implementation for base64 over should work. The rest of the guava uses are simple preconditions and some other checks that can easily be replaced with something else.

cyberluke:
Hi, I just currently use my fork with this pull request in our project and have no problems. What is the issue with base64 please? What does it affect?

Java 6 is unsupported, Java 8 is also unsupported now, hehe. The only supported is Java 10 and Java 9 for a short while.

So if I use jitpack.io like in your SDK. How can I use the specific pull request of yours that would solve the issue too?

The whole Google concept of having one Guava for Android and one Guava for the rest of Java world is definitely bad example of software engineering. But Google is monopoly, don't want to elaborate about goodwill and companies now. But this approach is bad and you solve only their amateur work now :-D. The real deal, the real Java is multiplatform :-D

Thank you

overcat:
We removed Guava in #524, so let's close this PR. Thank you for your contribution.

Issue #176, opened 2019-02-20
Allow custom base fee for transaction


tomerweller:
Right now the `TransactionBuilder` uses a hard-coded base fee that can't be modified. 
https://github.com/stellar/java-stellar-sdk/blob/a6fa9863556b9088d026ede113491335f400d4cf/src/main/java/org/stellar/sdk/Transaction.java#L371

PR #177, opened 2019-02-20
enable setting custom base fee on transactions as default and on builder

### LINKED ISSUE #171: how to set transaction fee?
builder() function in class Transaction create new transaction with fee:
`operations.length * BASE_FEE`

` public Transaction build() {
      Operation[] operations = new Operation[mOperations.size()];
      operations = mOperations.toArray(operations);
      Transaction transaction = new Transaction(mSourceAccount.getKeypair(), operations.length * BASE_FEE, mSourceAccount.getIncrementedSequenceNumber(), operations, mMemo, mTimeBounds);
      // Increment sequence number when there were no exceptions when creating a transaction
      mSourceAccount.incrementSequenceNumber();
      return transaction;
    }`

so we cannot change transaction fee?

nikhilsaraf:
changes:
 - allow setting a default base fee for all transactions: `Transaction.setDefaultBaseFee(200);`
 - allow setting a base fee on a per transaction basis: `txBuilder.setBaseFee(200);`

both functions will throw an IllegalArgumentException if you pass in a base fee that is _less than_ `100`.

tests included.

closes https://github.com/stellar/java-stellar-sdk/issues/176
EDIT @bartekn: close #171

### DIFF
--- a/src/main/java/org/stellar/sdk/Transaction.java
+++ b/src/main/java/org/stellar/sdk/Transaction.java
@@ -19,8 +19,8 @@ import static com.google.common.base.Preconditions.checkNotNull;
  * Represents <a href="https://www.stellar.org/developers/learn/concepts/transactions.html" target="_blank">Transaction</a> in Stellar network.
  */
 public class Transaction {
-  private static final int MINIMUM_BASE_FEE = 100;
-  private static int BASE_FEE = MINIMUM_BASE_FEE;
+  private static final int BASE_FEE = 100;
+  private static Integer defaultOperationFee;
 
   private final int mFee;
   private final KeyPair mSourceAccount;
@@ -52,12 +52,12 @@ public class Transaction {
     mSignatures.add(signer.signDecorated(txHash));
   }
 
-  public static void setDefaultBaseFee(int baseFee) {
-    if (baseFee < MINIMUM_BASE_FEE) {
-      throw new IllegalArgumentException("Invalid baseFee value: " + baseFee);
+  public static void setDefaultOperationFee(int opFee) {
+    if (opFee < BASE_FEE) {
+      throw new IllegalArgumentException("DefaultOperationFee cannot be smaller than the BASE_FEE (" + BASE_FEE + "): " + opFee);
     }
 
-    BASE_FEE = baseFee;
+    defaultOperationFee = opFee;
   }
 
   /**
@@ -267,7 +267,7 @@ public class Transaction {
     private TimeBounds mTimeBounds;
     List<Operation> mOperations;
     private boolean timeoutSet;
-    private int baseFee;
+    private Integer operationFee;
 
     public static final long TIMEOUT_INFINITE = 0;
 
@@ -281,7 +281,7 @@ public class Transaction {
       checkNotNull(sourceAccount, "sourceAccount cannot be null");
       mSourceAccount = sourceAccount;
       mOperations = Collections.synchronizedList(new ArrayList<Operation>());
-      baseFee = BASE_FEE;
+      operationFee = defaultOperationFee;
     }
 
     public int getOperationsCount() {
@@ -368,12 +368,12 @@ public class Transaction {
       return this;
     }
 
-    public Builder setBaseFee(int baseFee) {
-      if (baseFee < MINIMUM_BASE_FEE) {
-        throw new IllegalArgumentException("Invalid baseFee value: " + baseFee);
+    public Builder setOperationFee(int operationFee) {
+      if (operationFee < BASE_FEE) {
+        throw new IllegalArgumentException("OperationFee cannot be smaller than the BASE_FEE (" + BASE_FEE + "): " + operationFee);
       }
 
-      this.baseFee = baseFee;
+      this.operationFee = operationFee;
       return this;
     }
 
@@ -386,9 +386,14 @@ public class Transaction {
         throw new RuntimeException("TimeBounds has to be set or you must call setTimeout(TIMEOUT_INFINITE).");
       }
 
+      if (operationFee == null) {
+        System.out.println("[TransactionBuilder] The `operationFee` parameter of `Transaction` or `TransactionBuilder` is required. Setting to BASE_FEE=" + BASE_FEE + ". Future versions of this library will error if not provided.");
+        operationFee = BASE_FEE;
+      }
+
       Operation[] operations = new Operation[mOperations.size()];
       operations = mOperations.toArray(operations);
-      Transaction transaction = new Transaction(mSourceAccount.getKeypair(), operations.length * baseFee, mSourceAccount.getIncrementedSequenceNumber(), operations, mMemo, mTimeBounds);
+      Transaction transaction = new Transaction(mSourceAccount.getKeypair(), operations.length * operationFee, mSourceAccount.getIncrementedSequenceNumber(), operations, mMemo, mTimeBounds);
       // Increment sequence number when there were no exceptions when creating a transaction
       mSourceAccount.incrementSequenceNumber();
       return transaction;
--- a/src/main/java/org/stellar/sdk/Transaction.java
+++ b/src/main/java/org/stellar/sdk/Transaction.java
@@ -19,9 +19,6 @@ import static com.google.common.base.Preconditions.checkNotNull;
  * Represents <a href="https://www.stellar.org/developers/learn/concepts/transactions.html" target="_blank">Transaction</a> in Stellar network.
  */
 public class Transaction {
-  private static final int BASE_FEE = 100;
-  private static Integer defaultOperationFee;
-
   private final int mFee;
   private final KeyPair mSourceAccount;
   private final long mSequenceNumber;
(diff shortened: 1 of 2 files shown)


src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -51,6 +52,14 @@ public void sign(KeyPair signer) {
     mSignatures.add(signer.signDecorated(txHash));
   }
 
+  public static void setDefaultBaseFee(int baseFee) {
```
tomquisel: I might rename this to `setDefaultFee`, since the [base fee](https://www.stellar.org/developers/guides/concepts/fees.html) is fixed at 100 stroops (at least for now). I assume this is meant to set the fee that will be used for transactions submitted in this session, and could easily be set to something other than 100 stroops even if the network base fee doesn't change.

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -258,6 +267,7 @@ public String toEnvelopeXdrBase64() {
     private TimeBounds mTimeBounds;
     List<Operation> mOperations;
     private boolean timeoutSet;
+    private int baseFee;
```
tomquisel: Same thought as above (and this should apply in other place too). remove `base` so it's just called `fee`.

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -357,6 +368,15 @@ public Builder setTimeout(long timeout) {
       return this;
     }
 
+    public Builder setBaseFee(int baseFee) {
+      if (baseFee < MINIMUM_BASE_FEE) {
```
tomquisel: And  then I think `MINIMUM_BASE_FEE` should probably just be `BASE_FEE`. Also ideally it could be fetched from the last ledger like in the [js change](https://github.com/stellar/js-stellar-sdk/pull/234).

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -368,7 +388,7 @@ public Transaction build() {
 
       Operation[] operations = new Operation[mOperations.size()];
       operations = mOperations.toArray(operations);
-      Transaction transaction = new Transaction(mSourceAccount.getKeypair(), operations.length * BASE_FEE, mSourceAccount.getIncrementedSequenceNumber(), operations, mMemo, mTimeBounds);
+      Transaction transaction = new Transaction(mSourceAccount.getKeypair(), operations.length * baseFee, mSourceAccount.getIncrementedSequenceNumber(), operations, mMemo, mTimeBounds);
```
tomquisel: It'd be fantastic if it printed a deprecation warning (like was [added](https://github.com/stellar/js-stellar-base/pull/167) to the JS SDK) when `baseFee` hasn't been set explicitly

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -19,7 +19,8 @@
  * Represents <a href="https://www.stellar.org/developers/learn/concepts/transactions.html" target="_blank">Transaction</a> in Stellar network.
  */
 public class Transaction {
-  private static final int BASE_FEE = 100;
+  private static final int MINIMUM_BASE_FEE = 100;
+  private static int BASE_FEE = MINIMUM_BASE_FEE;
```
bartekn: It's a [convention](https://www.oracle.com/technetwork/articles/javaee/codeconventions-135099.html) to use uppercase letters for constants.

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -51,6 +52,14 @@ public void sign(KeyPair signer) {
     mSignatures.add(signer.signDecorated(txHash));
   }
 
+  public static void setDefaultBaseFee(int baseFee) {
+    if (baseFee < MINIMUM_BASE_FEE) {
+      throw new IllegalArgumentException("Invalid baseFee value: " + baseFee);
```
bartekn: Can it say that it's invalid because it's smaller than minimum?

src/main/java/org/stellar/sdk/Transaction.java
bartekn: We should actually change it to something like `perOperationFee` as it's multiplied by number of operations below.

src/main/java/org/stellar/sdk/Transaction.java
nikhilsaraf: yeah, my bad. will fix, thanks!

src/main/java/org/stellar/sdk/Transaction.java
nikhilsaraf: yes, good point, will rename. thanks.

src/main/java/org/stellar/sdk/Transaction.java
nikhilsaraf: yes, will update.

src/main/java/org/stellar/sdk/Transaction.java
nikhilsaraf: ok, will use perOperationFee so it's more descriptive.

src/main/java/org/stellar/sdk/Transaction.java
nikhilsaraf: will rename to `setDefaultOperationFee` to keep it more descriptive (bartek's comment below)

src/main/java/org/stellar/sdk/Transaction.java
nikhilsaraf: will do.

src/main/java/org/stellar/sdk/Transaction.java
nikhilsaraf: moved this issue to https://github.com/stellar/java-stellar-sdk/issues/181 as discussed.

nikhilsaraf:
@tomquisel @bartekn made all the requested changes, let me know if this is good to go. thanks!

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -20,6 +20,7 @@
  */
 public class Transaction {
   private static final int BASE_FEE = 100;
+  private static Integer defaultOperationFee;
```
westonal: Can't we avoid the null checks and initialize this value:
```
private static int defaultOperationFee = BASE_FEE;
```
Otherwise isn't this a breaking change forcing code to change to specify the operation fee.

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -366,9 +386,14 @@ public Transaction build() {
         throw new RuntimeException("TimeBounds has to be set or you must call setTimeout(TIMEOUT_INFINITE).");
       }
 
+      if (operationFee == null) {
+        System.out.println("[TransactionBuilder] The `operationFee` parameter of `Transaction` or `TransactionBuilder` is required. Setting to BASE_FEE=" + BASE_FEE + ". Future versions of this library will error if not provided.");
+        operationFee = BASE_FEE;
+      }
```
westonal: This can go if you initialize `defaultOperationFee`.

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -258,6 +267,7 @@ public String toEnvelopeXdrBase64() {
     private TimeBounds mTimeBounds;
     List<Operation> mOperations;
     private boolean timeoutSet;
+    private Integer operationFee;
```
westonal: This can be `int` if you initialize `defaultOperationFee`.

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -51,6 +52,14 @@ public void sign(KeyPair signer) {
     mSignatures.add(signer.signDecorated(txHash));
   }
 
+  public static void setDefaultOperationFee(int opFee) {
+    if (opFee < BASE_FEE) {
+      throw new IllegalArgumentException("DefaultOperationFee cannot be smaller than the BASE_FEE (" + BASE_FEE + "): " + opFee);
+    }
+
+    defaultOperationFee = opFee;
+  }
```
bartekn: Wondering if we should move `BASE_FEE`, `defaultOperationFee` and `setDefaultOperationFee` to `Transaction.Builder`. Looks like it's not used in `Transaction` directly.

nikhilsaraf: @bartekn transactionBuilder has this already. Instead of setting it on each transactionBuilder instance you can set it as a default for your whole session on Transaction if you want (as a consumer of the sdk). Happy to remove if you feel we should not have this. See the new tests in TransactionTest for examples.

bartekn: My point was to have a `static` field on `Transaction.Builder` instead of `Transaction`.

nikhilsaraf: Ah I see — will make the change and update this PR

Issue #178, opened 2019-02-20
dynamic fee retrieval based on operation fee stats endpoint


nikhilsaraf:
see similar issue in js sdk: https://github.com/stellar/js-stellar-sdk/issues/233
similar issue for allowing custom fees for transactions in java sdk: https://github.com/stellar/java-stellar-sdk/issues/176

nikhilsaraf:
closed by #180

PR #179, opened 2019-02-21
add test for operation fee stats against horizon testnet

nikhilsaraf:
functionality for operation fee stats already existed, added a test agains the testnet server

A mock test already exists, so we don't really need this but may be useful to have this test in the short term to ensure we're fetching the correct values

closes https://github.com/stellar/java-stellar-sdk/issues/178


bartekn:
In general tests shouldn't make network connections. Could you add `@Ignore` tag to it and people can uncomment when they want to run this specific test?

tomquisel:
Good point @bartekn. Even better, it seems like this file has other test cases that mock the web server. Could you use that approach maybe?

nikhilsaraf:
closing as discussed.

PR #182, opened 2019-02-27
solution to okhttpclient not shutting down cleanly

jillesvangurp:
- makes Server closable to workaround the issue that okhttpclient does not shut down cleanly.
- adds a second constructor to Server so users can provide their own okhttpclient configuration if they want to

I ran into this today with 0.6.0; seems to be a new issue relative to a 0.4.0. Closing okhttpclient by calling evictAll() on the connection pool fixes things. They may provide a proper solution at some point; in that case we should update the close method on Server.

Note. I assume we want to stick with java 1.6/Android compatibility. If not, we could modify this to AutoCloseable. This works a bit nicer in combination with try with resources introduced in 1.7.

Also see, https://github.com/square/okhttp/issues/3372


jillesvangurp:
@bartekn could we merge this now?

Issue #175, opened 2019-01-28
crashe in SSEStream when network is lost


dabitdev:
#169  this was closed and the crashes were found in 0.4.1
nevermind, If I open a stream and the device is moved to airplane mode these following crash is always observed.

```
  Process: com.blockeq.stellarwallet, PID: 4665
    java.lang.IllegalStateException: Failed 200
        at org.stellar.sdk.requests.SSEStream$StellarEventSourceListener.onFailure(SSEStream.java:176)
        at shadow.okhttp3.internal.sse.RealEventSource.processResponse(RealEventSource.java:83)
        at shadow.okhttp3.internal.sse.RealEventSource.onResponse(RealEventSource.java:54)
        at shadow.okhttp3.RealCall$AsyncCall.execute(RealCall.java:153)
        at shadow.okhttp3.internal.NamedRunnable.run(NamedRunnable.java:32)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at java.lang.Thread.run(Thread.java:764)
     Caused by: javax.net.ssl.SSLException: Read error: ssl=0x71aac2d788: I/O error during system call, Connection timed out
        at com.android.org.conscrypt.NativeCrypto.SSL_read(Native Method)
        at com.android.org.conscrypt.NativeSsl.read(NativeSsl.java:399)
        at com.android.org.conscrypt.ConscryptFileDescriptorSocket$SSLInputStream.read(ConscryptFileDescriptorSocket.java:546)
        at shadow.okio.Okio$2.read(Okio.java:140)
        at shadow.okio.AsyncTimeout$2.read(AsyncTimeout.java:237)
        at shadow.okio.RealBufferedSource.request(RealBufferedSource.java:68)
        at shadow.okio.RealBufferedSource.require(RealBufferedSource.java:61)
        at shadow.okio.RealBufferedSource.readHexadecimalUnsignedLong(RealBufferedSource.java:304)
        at shadow.okhttp3.internal.http1.Http1Codec$ChunkedSource.readChunkSize(Http1Codec.java:469)
        at shadow.okhttp3.internal.http1.Http1Codec$ChunkedSource.read(Http1Codec.java:449)
        at shadow.okio.RealBufferedSource.indexOfElement(RealBufferedSource.java:394)
[...]

PR #187, opened 2019-03-22
Add error listener

sonh:
Need to handle the error occurs when lost connection.

### DIFF
--- a/src/main/java/org/stellar/sdk/requests/EventListener.java
+++ b/src/main/java/org/stellar/sdk/requests/EventListener.java
@@ -9,4 +9,6 @@ public interface EventListener<T> {
    * @param object object deserialized from the event data
    */
   void onEvent(T object);
+
+  void onError();
 }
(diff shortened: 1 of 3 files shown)


src/main/java/org/stellar/sdk/requests/EventListener.java
```diff
@@ -9,4 +9,6 @@
    * @param object object deserialized from the event data
    */
   void onEvent(T object);
+
+  void onError();
```
jillesvangurp: pass the exception as a parameter; probably a common use case for onError will be logging

src/main/java/org/stellar/sdk/requests/SSEStream.java
```diff
@@ -172,11 +172,14 @@ public void onFailure(EventSource eventSource, @Nullable Throwable t, @Nullable
       if (t != null) {
         if (t instanceof SocketException) {
           // not a failure, server disconnected
+          listener.onError();
```
jillesvangurp: we probably should have custom callbacks for expected events like this

jillesvangurp:
StreamingSmokeTest is broken. Otherwise liking the idea of callbacks instead of throwing exceptions that are swallowed. I do think we should have a couple of different callbacks for expected situations like the server disconnecting after a while.

sonhvp:
#175
Currently, we will get a crash if lost connection. Instead of throwing exceptions, we should have some callbacks to handle exceptions.

PR #194, opened 2019-04-17
Add Support for InflationOperation in Operation.fromXdr

jakemayeux:
Attempting to decode an inflation operation using `Transaction.fromEnvelopeXdr` would result in an "Unknown operation body INFLATION" exception. 

This change adds a case for this to support decoding Inflation operations from XDR.


jmayeux-circle:
@nullstyle Any chance you could do a point release for this sometime soon?

nullstyle:
@jmayeux-circle I plan on doing a point release within the next few weeks.

PR #202, opened 2019-05-07
feat: Set network when building a transaction

### LINKED ISSUE #155: Is it possible to concurrently use more than one network configuration?
Hi, i am trying to figure out if i could use multiple network configurations, concurrently. Our application works with both a private network and the Stellar public network. It would be really great if we could switch dynamically between those different configurations on demand, in a way that works with concurrent requests. Any ideas of how something like this could work in the existing implementation?

overcat:
resolve #155

### DIFF
--- a/src/main/java/org/stellar/sdk/Network.java
+++ b/src/main/java/org/stellar/sdk/Network.java
@@ -37,7 +37,7 @@ public class Network {
      * Returns network id (SHA-256 hashed <code>networkPassphrase</code>).
      */
     public byte[] getNetworkId() {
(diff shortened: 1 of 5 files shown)


jillesvangurp:
I think this needs a bit of rethinking and a different approach. It doesn't make sense to configure the network per transaction while talking to the same server endpoint. All requests to that server are implicitly for the network that server is running.

The real problem here is that Network keeps a weird global variable `private static Network current;`

A better approach would be to make this a constructor injected property for the Server.

PR #200, opened 2019-05-05
bugfix: orderbook stream

overcat:
resolve #196

PR #203, opened 2019-05-14
0.7.0

bartekn:
To be included:
- [x] #206
- [x] #205

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,17 @@ As this project is pre 1.0, breaking changes may happen for minor version bumps.
 
 ### Deprecations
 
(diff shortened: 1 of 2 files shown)


overcat:
Can we include #200 and #201 in 0.7.0?

bartekn:
#201 is good. For #200 I think we need a different fix (if I understand correctly there's no paging token for order book response?). Silencing exceptions is generally bad idea.

overcat:
Hi @bartekn, there's no paging token for order book response, so we get `UnsupportedOperationException` from [here](https://github.com/stellar/java-stellar-sdk/blob/98da7754299320a777bbad2485c69911c13dc99f/src/main/java/org/stellar/sdk/responses/Response.java#L34), can you show me how to fix it gracefully?

bartekn:
@overcat added a PR fixing this with a different approach here: #206.

PR #211, opened 2019-06-05
Set source compatibility to Java 1.8

tamirms:
Java 8 was released in 2014 and introduces a lot of nice features such as lambdas, streams,
and completable futures. By upgrading the Java Stellar SDK to use Java 8 we will be
able to take advantage of these features to improve the code quality of the Stellar client.

Issue #184, opened 2019-02-28
TimeBounds has to be set or you must call setTimeout(TIMEOUT_INFINITE)


prettyCoders:
It throw a exception when i start building the transaction use java sdk!

rolivieri:
I just ran into this same issue as I was following the instructions here (java sdk): https://www.stellar.org/developers/guides/get-started/transactions.html

Then found this issue (js sdk): https://github.com/stellar/docs/issues/431. The same fix is needed for the sample Java code that is provided in the instructions. Invoking the `setTimeout()`  on the transaction builder object fixes this error.

PR #212, opened 2019-06-05
 Add test which checks that a Transaction must have a timeout

tamirms:
When investigating https://github.com/stellar/java-stellar-sdk/issues/184 I realized that there was no unit test which tested that a timeout is required to build a Transaction object.

Also, I decided to use JUnit5 so that we can easily test that certain expressions throw exceptions in a clean manner.

depends on https://github.com/stellar/java-stellar-sdk/pull/211

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,6 +38,10 @@ repositories {
     mavenCentral()
 }
 
+test {
+    useJUnitPlatform()
+}
+
 dependencies {
     compile "com.squareup.okhttp3:okhttp:${okhttpclientVersion}"
     compile "com.squareup.okhttp3:okhttp-sse:${okhttpclientVersion}"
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -51,9 +55,15 @@ dependencies {
     testCompile 'org.mockito:mockito-core:2.21.0'
     testCompile "com.squareup.okhttp3:mockwebserver:${okhttpclientVersion}"
     testCompile 'junit:junit:4.12'
+    testRuntime("org.junit.vintage:junit-vintage-engine:5.2.0")
```
jillesvangurp: I think having both junit 4 and 5 on the classpath is not a good idea. I'd be in favor of upgrading everything to junit 5 but that's a bigger PR. Probably very straightforward to do.

build.gradle
```diff
@@ -15,7 +15,7 @@ apply plugin: 'maven' // needed to add the install task for jitpack.io
 apply plugin: 'com.github.ben-manes.versions' // gradle dependencyUpdates -Drevision=release
 apply plugin: 'project-report' // gradle htmlDependencyReport
 
-sourceCompatibility = 1.6
+sourceCompatibility = 1.8
```
jillesvangurp: there's a separate PR for this.

build.gradle
tamirms: thanks, I created an issue for this https://github.com/stellar/java-stellar-sdk/issues/215

Issue #98, opened 2018-05-13
Implement equals method on Keypair class


basmoorkens:
In a java sample application I'm writing I am pulling transactions from horizon. I want to check these transactions that the recipient Keypair is equal to a Keypair I created in my code see the code below:

I create my Keypair like this:
`KeyPair myPair =  KeyPair.fromSecretSeed("myseed");`

I get the Keypair from the TX like this: 
`KeyPair txKeypair = ((PaymentOperationResponse) payment).getTo();`

When I examine the byte[] from the Keypair method getPublicKey() I see that both KeyPairs have the same byte array so for me the accounts are logically the same. Is there a reason no compare method is in the Keypair object?

For now I worked around this issue by comparing the 2 byte[]

PR #221, opened 2019-06-11
Implement hashCode and equals on Stellar classes

tamirms:
fixes https://github.com/stellar/java-stellar-sdk/issues/76 and https://github.com/stellar/java-stellar-sdk/issues/98

### DIFF
--- a/src/main/java/org/stellar/sdk/xdr/Uint256.java
+++ b/src/main/java/org/stellar/sdk/xdr/Uint256.java
@@ -33,9 +33,7 @@ public class Uint256  {
   }
 
   public int hashCode() {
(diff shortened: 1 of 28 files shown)


src/main/java/org/stellar/sdk/xdr/Uint256.java
```diff
@@ -30,4 +31,21 @@ public static Uint256 decode(XdrDataInputStream stream) throws IOException {
   stream.read(decodedUint256.uint256, 0, uint256size);
     return decodedUint256;
   }
+
+  public int hashCode() {
+    return Arrays.hashCode(
```
ire-and-curses: Could be one line

PR #222, opened 2019-06-12
Implement hashCode and equals on Stellar classes

tamirms:
In https://github.com/stellar/java-stellar-sdk/pull/221 , I realized that I was including hashCode and equals implementations in the files which were generated by `xdrgen`. 

It's not a good idea to mix handwritten code with generated code. I added a patch `xdrgen` https://github.com/stellar/xdrgen/pull/46 to include hashCode and equals implementations in the code it generates.

The latest commit in this PR includes the XDR classes from my patched version of `xdrgen`.


src/main/java/org/stellar/sdk/xdr/AccountMergeResult.java
```diff
@@ -58,4 +60,17 @@ public static AccountMergeResult decode(XdrDataInputStream stream) throws IOExce
...
+    return Objects.hash(this.code, this.sourceAccountBalance);
+  }
+  @Override
+  public boolean equals(Object object) {
+    if (object == null || !(object instanceof AccountMergeResult)) {
+      return false;
+    }
+
+    AccountMergeResult other = (AccountMergeResult) object;
+    return Objects.equals(this.sourceAccountBalance, other.sourceAccountBalance) && Objects.equals(this.code, other.code);
```
ire-and-curses: It seems surprising that this generated code is coming up with a different order for the equals vs the arguments  in the hashCode. Other generated files I randomly sampled seem to have the same order.

bartekn:
It looks good. I have only one question/concern. In objects with a single property like `MemoId` you calculate the hash using `Objects.hash(...)`, like:
```java
Objects.hash(this.id);
```
However, there's a warning in [`Objects.hash`](https://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#hash(java.lang.Object...)) documentation:
> Warning: When a single object reference is supplied, the returned value does not equal the hash code of that object reference. This value can be computed by calling hashCode(Object).

I'm wondering if it has any implications but I can't find any example why this could be wrong. Do you?

tamirms:
@bartekn good catch. I found this on stack overflow

https://stackoverflow.com/a/11656642

If you look at the comments they link to the implementation of the function

http://hg.openjdk.java.net/jdk10/jdk10/jdk/file/777356696811/src/java.base/share/classes/java/util/Arrays.java#l4677

it seems like `Objects.hash(this.id)` will return 31  +  `this.id.hashCode()`

I will amend my xdrgen PR to handle hash codes for objects with a single field and then I'll update this PR again

src/main/java/org/stellar/sdk/xdr/AccountMergeResult.java
tamirms: @ire-and-curses I pushed a commit to https://github.com/stellar/xdrgen/pull/46 which fixes this ordering issue and the `Objects.hash()` issue that @bartekn  found

PR #224, opened 2019-06-13
Remove Network singleton

tamirms:
Instead of using a static global variable to determine the network for a transaction, each
Server will be bound to a specific network. A transaction will also be bound to a network.
If you try to submit a transaction to a server which is bound to a different network an
exception will be raised.

fixes https://github.com/stellar/java-stellar-sdk/issues/155

@jillesvangurp I'm using the approach you outlined in the issue above ^


src/main/java/org/stellar/sdk/Network.java
```diff
@@ -26,46 +28,30 @@ public Network(String networkPassphrase) {
...
     }
 
-    /**
-     * Returns currently used Network object.
-     */
-    public static Network current() {
-        return current;
+    @Override
+    public int hashCode() {
+        return Objects.hash(this.networkPassphrase);
```
bartekn: Use `hashCode`.

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -376,6 +395,14 @@ public Builder setOperationFee(int operationFee) {
       return this;
     }
 
+    public Builder setNetwork(Network network) {
+      if (mNetwork != null) {
+        throw new RuntimeException("Network has already been configured.");
+      }
+      mNetwork = checkNotNull(network, "Network cannot be null");
+      return this;
+    }
```
bartekn: Do you think it's also worth adding this as a constructor param? This must always be set, as `sourceAccount`.

src/main/java/org/stellar/sdk/Server.java
```diff
@@ -28,28 +29,36 @@
...
         );
 
     }
 
-    public Server(String serverURI, OkHttpClient httpClient, OkHttpClient submitHttpClient) {
+    public Server(
+            String serverURI,
+            OkHttpClient httpClient,
+            OkHttpClient submitHttpClient,
+            Network network
```
bartekn: I'm wondering if, instead, we should request `/` (`Server.root()`) that exposes `network_passphrase` value. Not in a constructor but, ex. once in `submitTransaction`. It's simpler and prevents situations where users mistakenly set a network passphrase to a wrong one and receive bad signature errors for their transactions.

jillesvangurp:
Thanks for picking this up; this is one of those things I never got around to tackling. Looks good to me as is. 

It does start to look like we might want some kind of builder for the Server class as a follow up to this PR. On the Kotlin side, I can avoid these by using default arguments.

src/main/java/org/stellar/sdk/Server.java
jillesvangurp: @bartekn that could potentially be interesting to make the server instance self configuring where it initializes by doing a request to Server.root(). That could make setting up a standalone network simpler.

src/main/java/org/stellar/sdk/Server.java
```diff
@@ -191,8 +193,12 @@ public TransactionsRequestBuilder transactions() {
      * @throws IOException
      */
     public SubmitTransactionResponse submitTransaction(Transaction transaction) throws IOException {
-        if (!this.network.equals(transaction.getNetwork())) {
-            throw new NetworkMismatchException(this.network, transaction.getNetwork());
+        if (!this.network.isPresent()) {
+            this.root();
+        }
```
bartekn: There's a possible race condition when multiple `submitTransaction` are executed at the same time.

tamirms: @bartekn I believe this should be fixed in my latest commit

bartekn: It's better and it works but it's still possible that `root()` will be called twice when `getNetwork()` in the first line of `submitTransaction` returns `null` for 2 or more threads. It's not a big deal because calls to `root()` are cheap so up to you if you want to fix it.

tamirms: yeah, I think that's fine

PR #232, opened 2019-07-19
update dependencies (okhttpclient, guava)

### LINKED ISSUE #158: Upgrade okhttp to 4.0.1

jillesvangurp:
The big change is okhttpclient 4 which has been rewritten in Kotlin while remaining compatible, this fixes #158

the other run time dependency that updated is gauva.

Both libraries should be fine for Android.

I also updated gradle to the latest version as well as a few plugins and test dependencies


jillesvangurp:
I'll close this as there has been no action on this and by now everything probably needs updating again. 

With 14 open pull requests, I recommend taking action on most of them even if it is just to close them.

PR #241, opened 2019-09-27
Upgrade to protocol v12

tamirms:
Fixes https://github.com/stellar/java-stellar-sdk/issues/240

* Add strict receive / send operation and response classes
* Keep original  PathPaymentOperation and PathPaymentOperationResponse classes but add deprecation warnings
* Include `path` response field in path payment operation response classes


src/main/java/org/stellar/sdk/PathPaymentStrictSendOperation.java
```diff
@@ -0,0 +1,225 @@
...
+import org.stellar.sdk.xdr.OperationType;
+import org.stellar.sdk.xdr.PathPaymentStrictSendOp;
+
+import java.util.Arrays;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Represents <a href="https://www.stellar.org/developers/learn/concepts/list-of-operations.html#path-payment" target="_blank">PathPaymentStrictSend</a> operation.
```
bartekn: The doc is not updated it but I believe the URL will looks like this:
```suggestion
 * Represents <a href="https://www.stellar.org/developers/learn/concepts/list-of-operations.html#path-payment-strict-send" target="_blank">PathPaymentStrictSend</a> operation.
```

xdr/Stellar-transaction.x
```diff
@@ -784,6 +849,8 @@ case opINNER:
         BumpSequenceResult bumpSeqResult;
     case MANAGE_BUY_OFFER:
 	ManageBuyOfferResult manageBuyOfferResult;
+    case PATH_PAYMENT_STRICT_SEND:
+        PathPaymentStrictSendResult pathPaymentStrictSendResult;
```
bartekn: I think there were more changes in XDR files besides this one, see: https://github.com/stellar/go/commit/50b0a8d71a159659a992e72f666900fab365734e#diff-c99bb2d3a1bde548287069d1fe768b58 Can you update all of them?

src/main/java/org/stellar/sdk/responses/OperationDeserializer.java
```diff
@@ -7,45 +7,56 @@
...
 
 import java.lang.reflect.Type;
 
 class OperationDeserializer implements JsonDeserializer<OperationResponse> {
+  private static final OperationType[] AllOperationTypes = OperationType.values();
   @Override
   public OperationResponse deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context) throws JsonParseException {
     // Create new Gson object with adapters needed in Operation
     Gson gson = new GsonBuilder()
+            .registerTypeAdapter(Asset.class, new AssetDeserializer())
```
bartekn: Good catch!

src/main/java/org/stellar/sdk/PathPaymentStrictReceiveOperation.java
```diff
@@ -0,0 +1,225 @@
...
+import org.stellar.sdk.xdr.OperationType;
+import org.stellar.sdk.xdr.PathPaymentStrictReceiveOp;
+
+import java.util.Arrays;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Represents <a href="https://www.stellar.org/developers/learn/concepts/list-of-operations.html#path-payment" target="_blank">PathPaymentStrictReceive</a> operation.
```
ire-and-curses: Link will likely be `path-payment-strict-receive`

xdr/Stellar-transaction.x
tamirms: I double checked and all the .x files seem up to date. I think the XDR files in the go mono repo were previously out of date and that's why the go PR include changes to several XDR files.

If you look at https://github.com/stellar/stellar-core/commits/master/src/xdr you will find that the most recent commits to the xdr directory in stellar core involve changes to only Stellar-transaction.x

Issue #236, opened 2019-08-21
add support for querying operation endpoint with 'join' parameter


tamirms:
Horizon v0.19 introduces a join parameter when querying for operations. This allows users to get transaction details with operation responses.

ire-and-curses:
Go implementation: [go:1521](https://github.com/stellar/go/issues/1521)

tamirms:
fixed in https://github.com/stellar/java-stellar-sdk/pull/242

PR #242, opened 2019-09-30
Allow operations to include transaction data

tamirms:
Horizon v0.19 introduces a join parameter when querying for operations. This allows users to get transaction details with operation responses.

fixes https://github.com/stellar/java-stellar-sdk/issues/236

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Deprecated | New method/class
 
 * Stellar Protocol 12 compatibility.
 * Include `path` property in path payment operation responses.
+* Provide method for constructing operations requests which include transaction data in the operations response.
 
 ## 0.9.0
 * Use strings to represent account ids instead of KeyPair instances because account ids will not necessarily be valid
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Deprecated | New method/class
 
 * Stellar Protocol 12 compatibility.
 * Include `path` property in path payment operation responses.
+* Provide method for constructing operations requests which include transaction data in the operations response.
```
bartekn: Can you add a method name?

src/main/java/org/stellar/sdk/requests/OperationsRequestBuilder.java
```diff
@@ -9,15 +11,19 @@
...
 
 import java.io.IOException;
+import java.util.Set;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
  * Builds requests connected to operations.
  */
 public class OperationsRequestBuilder extends RequestBuilder {
```
bartekn: We need the same thing for `PaymentsRequestBuilder`.

Issue #235, opened 2019-08-02
Support for SEP 10 challenge transaction


ire-and-curses:
SEP 10 describes a [specific form](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#response) for the "challenge" transaction used for authentication. Providing a helper at the SDK level makes sense to save users from reimplementing the spec.

Here are example implementations in [Go](https://github.com/stellar/go/pull/1468) and [JS](https://github.com/stellar/js-stellar-sdk/pull/375).

dabitdev:
Is there any equivalent to the `WebAuthenticator` form the ios SDK?

dabitdev:
my fault > Sep10Challenge.class

leighmcculloch:
Closed by #264.

PR #243, opened 2019-09-30
Add helper method to generate SEP10 challenge

tamirms:
fixes https://github.com/stellar/java-stellar-sdk/issues/235

### DIFF
--- a/src/main/java/org/stellar/sdk/Sep10Challenge.java
+++ b/src/main/java/org/stellar/sdk/Sep10Challenge.java
@@ -1,0 +1,39 @@
+package org.stellar.sdk;
+
+import java.security.SecureRandom;
+
+public class Sep10Challenge {
+  /**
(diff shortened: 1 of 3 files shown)


src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -0,0 +1,39 @@
...
+  public static String newChallenge(
+      KeyPair signer,
+      Network network,
+      String clientAccountId,
+      String anchorName,
+      TimeBounds timebounds
+  ) {
+    byte[] nonce = new byte[64];
+    SecureRandom random = new SecureRandom();
+    random.nextBytes(nonce);
```
bartekn: You may want to check a discussion here: https://github.com/stellar/stellar-protocol/pull/366#discussion_r330134927. I don't fully agree with reasoning there but it looks like you need to generate 48 bytes and base64 encode them to fully comply with SEP-10.

Issue #257, opened 2019-12-05
Memo check doesn't take into account ASCII encoding


Inkon:
When [this transaction](https://testnet.steexp.com/tx/213d52180f5e74fc4e4bc2c86740a44f98a40ab31e6718cc032e95f9084ae535) was decoded from Json using SDK it failed with following Exception:
```
Caused by: org.stellar.sdk.MemoTooLongException: text must be <= 28 bytes. length=54
        at org.stellar.sdk.MemoText.<init>(MemoText.java:21)
```
As text memo could use either UTF8 or ASCII encoding second option should be checked as well, when decoding transactions from json.

leighmcculloch:
A memo could also use another encoding that isn't ASCII or UTF-8, so I'm unconvinced we should limit to decoding the data with just those two. Java String's are UTF-16, so unfortunately the moment we decode the memo from bytes into the String we're possibly changing the string.

Could we instead change how memos are decoded so that the `MemoText` class will accept as `byte[]` and will pass that through? The `MemoText can still support `String` too as an input, but will convert that into UTF-8 bytes at the time of input.

Inkon:
Yes, you are right and memo could use any encoding. Currently, there are 2 places where `MemoText` is used:

1. User input (somebody using SDK attaching memo to transaction) -- which we could validate and ask user to provide encoding. However, if I understand correctly (I haven't looked in go sources though), it must be encoded either in UTF-8 or ASCII (judging from docs) and would be rejected by Horizon if it's too long
2. Parsing transaction Json received from blockchain. In that scenario do we really need to validate the string at all? My implementation still validates the string but under assumption that string of memo in json is encoded in this 2 options.

So, regarding your comment.

> Java String's are UTF-16, so unfortunately the moment we decode the memo from bytes into the String we're possibly changing the string.
Yes, that's true, so for user input we can ask for byte string representation. (Q: how do we encoded it? Using UTF-8/ASCII? Using user-provided encoding?)

> Could we instead change how memos are decoded so that the MemoText class will accept as byte[] and will pass that through? 

But we still need to convert it to string to submit to Horizon, don't we? 

> The MemoText can still support String too as an input, but will convert that into UTF-8 bytes at the time of input.
 
Won't there be a problem with string that are too long in utf8 encoding but are short enough in ASCII encoding? (As in example transaction)

Thanks!

Inkon:
Another question why can't we simply use memo from XDR as ground truth on decoding stage?

[...]

PR #258, opened 2019-12-05
[#257] Fixed failures on decoding ASCII text memos

Inkon:
Added fixes for issue #257 only to JSON transaction deserializer, which is now get length as minimum of UTF8 and ASCII (under assumption that data from the network is correct)
All other places use old code, which was left untouched.

### DIFF
--- a/src/main/java/org/stellar/sdk/Memo.java
+++ b/src/main/java/org/stellar/sdk/Memo.java
@@ -2,6 +2,8 @@ package org.stellar.sdk;
 
 import com.google.common.io.BaseEncoding;
 
+import java.nio.charset.Charset;
+
 /**
  * <p>The memo contains optional extra information. It is the responsibility of the client to interpret this value. Memos can be one of the following types:</p>
  * <ul>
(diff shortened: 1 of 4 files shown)


src/main/java/org/stellar/sdk/Memo.java
```diff
@@ -24,10 +26,18 @@ public static MemoNone none() {
 
     /**
      * Creates new {@link MemoText} instance.
-     * @param text
+     * @param text test in utf8 encoding
```
tamirms: should this be "text in utf8 encoding" instead of "test in utf8 encoding?"

src/main/java/org/stellar/sdk/Memo.java
```diff
@@ -24,10 +26,18 @@ public static MemoNone none() {
...
      */
     public static MemoText text(String text) {
-        return new MemoText(text);
+        return new MemoText(text, Charset.forName("UTF8"));
+    }
+
+    /**
+     * Creates new {@link MemoText} instance.
+     * @param text test in unknown encoding
```
tamirms: should this be "text in unknown encoding" instead of "test in unknown encoding?"

Inkon:
I'm closing the PR as it's fixed in #259

PR #259, opened 2019-12-09
Use bytes to represent memo text

tamirms:
Fixes https://github.com/stellar/java-stellar-sdk/issues/257

The Java SDK assumes that all memo texts must be valid UTF 8 strings. However, that assumption is not valid. It turns out that any sequence of bytes is allowed to appear in the memo field (see https://github.com/stellar/go/issues/2022 ).

Therefore, the correct representation for the memo field is a byte array.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 As this project is pre 1.0, breaking changes may happen for minor version bumps. A breaking change will get clearly notified in this log.
 
+## 0.12.0
+
+* Represent memo text contents as bytes because a memo text may not be valid UTF-8 string.
+
 ## 0.11.0
 
 * Fix bug in `org.stellar.sdk.requests.OperationsRequestBuilder.operation(long operationId)`. The method submitted an HTTP request to Horizon with the following path, /operation/<id> , but the correct path is /operations/<id>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ As this project is pre 1.0, breaking changes may happen for minor version bumps.
 ## 0.12.0
 
 * Represent memo text contents as bytes because a memo text may not be valid UTF-8 string (https://github.com/stellar/java-stellar-sdk/issues/257).
+* Validate name length when constructing org.stellar.sdk.ManageDataOperation instances.
+* Validate home domain length when constructing org.stellar.sdk.SetOptionsOperation instances.
 
 ## 0.11.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ As this project is pre 1.0, breaking changes may happen for minor version bumps.
 
 ## 0.12.0
 
-* Represent memo text contents as bytes because a memo text may not be valid UTF-8 string.
+* Represent memo text contents as bytes because a memo text may not be valid UTF-8 string (https://github.com/stellar/java-stellar-sdk/issues/257).
 
 ## 0.11.0
 
--- a/src/main/java/org/stellar/sdk/ManageDataOperation.java
+++ b/src/main/java/org/stellar/sdk/ManageDataOperation.java
@@ -3,8 +3,6 @@ package org.stellar.sdk;
 import com.google.common.base.Objects;
 import org.stellar.sdk.xdr.*;
 
-import java.io.UnsupportedEncodingException;
-import java.nio.charset.Charset;
 import java.util.Arrays;
 
 import static com.google.common.base.Preconditions.checkNotNull;
--- a/src/main/java/org/stellar/sdk/ManageDataOperation.java
+++ b/src/main/java/org/stellar/sdk/ManageDataOperation.java
@@ -1,10 +1,7 @@
 package org.stellar.sdk;
 
 import com.google.common.base.Objects;
-import org.stellar.sdk.xdr.DataValue;
-import org.stellar.sdk.xdr.ManageDataOp;
-import org.stellar.sdk.xdr.OperationType;
-import org.stellar.sdk.xdr.String64;
+import org.stellar.sdk.xdr.*;
 
 import java.io.UnsupportedEncodingException;
 import java.nio.charset.Charset;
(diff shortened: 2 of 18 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 As this project is pre 1.0, breaking changes may happen for minor version bumps. A breaking change will get clearly notified in this log.
 
+## 0.12.0
+
+* Represent memo text contents as bytes because a memo text may not be valid UTF-8 string.
```
ire-and-curses: Let's link to the issue (#257) from the changelog

src/main/java/org/stellar/sdk/MemoText.java
```diff
@@ -4,26 +4,33 @@
...
+  public MemoText(byte[] text) {
+    this.text = checkNotNull(text, "text cannot be null");
+    if (this.text.length > 28) {
+      throw new MemoTooLongException("text must be <= 28 bytes. length=" + String.valueOf(this.text.length));
     }
   }
 
   public String getText() {
-    return text;
+    return new String(text, Charset.forName("UTF-8"));
```
ire-and-curses: Consider for consistency
```suggestion
    return new String(this.text, Charset.forName("UTF-8"));
```

src/main/java/org/stellar/sdk/xdr/Memo.java
```diff
@@ -15,7 +16,7 @@
 //  case MEMO_NONE:
 //      void;
 //  case MEMO_TEXT:
-//      string text<28>;
+//      opaque text<28>;
```
leighmcculloch: At the top of this file is a comment saying the file file is generated with [xdrgen], but I don't see any updates to xdrgen that would have caused this. Are these changes being made manually to this file and the .x file in this repository without changing xdrgen? We won't be able to use xdrgen to generate these files again without losing these changes.
https://github.com/stellar/java-stellar-sdk/blob/9739b4bd877e4c72809d1f7da7e56a250c0cd4d1/src/main/java/org/stellar/sdk/xdr/Memo.java#L1-L2

[xdrgen]: https://github.com/stellar/xdrgen

I think we should keep this as `string` and change xdrgen so that it internally stores the field as a `byte[]` and offers up both `getText() : String` and `getTextBytes() : byte[]` functions, and similar setters, which should be backwards compatible and a non-breaking change to code it generates.

tamirms: see the discussion for this issue: https://github.com/stellar/go/issues/2022

leighmcculloch: I DM'd with @tamirms and he confirmed this isn't a manual edit and the file generated by xdrgen, which was the main thing I was concerned about. I was misunderstanding the generated functions the byte type exports. Nice!

xdr/Stellar-transaction.x
```diff
@@ -335,7 +335,7 @@ union Memo switch (MemoType type)
 case MEMO_NONE:
     void;
 case MEMO_TEXT:
-    string text<28>;
+    opaque text<28>;
```
leighmcculloch: Where do we keep the source of truth for our .x files? Does this get updated there also so that all other SDKs also get updated?

tamirms: I think they are kept here https://github.com/stellar/stellar-core/tree/master/src/xdr

I'm waiting for confirmation in https://github.com/stellar/go/issues/2022 that we're going to change the type of the memo text field. Once we have confirmation we'll need to update the .x files in stellar core and in the go monorepo

bartekn: Files in `xdr` should be exactly the same as in stellar-core repo, we shouldn't change them here nor autogenerated files in: `src/test/java/org/stellar/sdk/xdr`.

tamirms: @bartekn 

> I'm waiting for confirmation in stellar/go#2022 that we're going to change the type of the memo text field. Once we have confirmation we'll need to update the .x files in stellar core and in the go monorepo

bartekn: If this is merged after the changes are in stellar-core master then 👍.

tamirms:
David wasn't in favor of changing the type of the memo field, https://github.com/stellar/go/issues/2022#issuecomment-564185795 . So I was left with no choice except for changing the implementation of how XDR strings were represented in the Java code generator: https://github.com/stellar/xdrgen/pull/48

I have removed the changes to xdr/Stellar-transaction.x and regenerated the java classes using the version of xdrgen from my pull request ( https://github.com/stellar/xdrgen/pull/48 ).

Please review https://github.com/stellar/xdrgen/pull/48 first and then review this PR

src/main/java/org/stellar/sdk/MemoText.java
```diff
@@ -2,53 +2,61 @@
...
+import java.util.Arrays;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
  * Represents MEMO_TEXT.
  */
 public class MemoText extends Memo {
-  private String text;
+  private byte[] text;
```
leighmcculloch: If `text` was an `XdrString` then we'd keep the UTF-8 specific conversion code in only one place, in the XdrString class. Right now that code is spread out across XdrString and line 19, and any other class like MemoText that we need to change will need to know to use UTF-8 too.

src/main/java/org/stellar/sdk/ManageDataOperation.java
```diff
@@ -20,6 +20,10 @@
   private ManageDataOperation(String name, byte[] value) {
     this.name = checkNotNull(name, "name cannot be null");
     this.value = value;
+
+    if (new XdrString(this.name).getBytes().length > 64) {
+      throw new IllegalArgumentException("name cannot exceed 64 bytes");
+    }
```
leighmcculloch: This change seems to be unrelated to the rest of this change. The rest of this change is specifically about memo text, but this change is to do with manage data operations. It seems out of scope and like it belongs in a separate PR.

tamirms: if I separated this change in a new PR it would depend on this branch because it requires `XdrString`. Lets keep both changes in the same PR for now. It shouldn't be too burdensome to review the extra changes

PR #264, opened 2020-01-30
feat: add support for SEP-0010.

### LINKED ISSUE #235: Support for SEP 10 challenge transaction

### LINKED ISSUE #263: Update challenge transaction helpers for SEP-10 v1.3.0
### Summary

A change has been released in SEP-10 Web Authentication v1.3.0 that adds support for:
 1. Accounts not using the master key as a signer.
 2. Accounts multiple signers.

SEP-10 v1.3.0 Change: stellar/stellar-protocol@ea0d7edc3bed84a24970c5fe5fc34993e6c97e48

The new functionality requires changes to any SDK implementing challenge verification.

### Details and Reference Implementations

The Stellar Go SDK is the first SDK we've implemented this new functionality and it can be used as a reference. It has implemented the following changes and similar changes may be required in this SDK:
 * Deprecated the existing verification function (`VerifyChallengeTx`).
 * Added new functions to:
   * Read the client account ID out of a challenge transaction without verifying client signatures, so that servers can lookup the signers of the account before verifying. (`ReadChallengeTx`)
   * Verify signatures on a challenge transaction meet a threshold. (`VerifyChallengeTxThreshold`)
   * Verify signatures on a challenge transaction match signers. (`VerifyChallengeTxSigners`)
 * Increased the timeout on the challenge transaction from 5 minutes to 15 minutes.

Go SDK Change: stellar/go@8ff0848c31a940ab76f1a9177af0d9dd405ffa9b

An example of a SEP-10 server using the v1.3.0 verification process is implemented here:
https://github.com/stellar/go/tree/master/exp/services/webauth

If anyone has any questions feel free to join us in `#dev-discussion` in `stellar.public` Keybase.

overcat:
This resolves #263 and resolves #235.

Feel free to add comments if you have any questions.

Thanks.

### DIFF
--- a/src/main/java/org/stellar/sdk/Sep10Challenge.java
+++ b/src/main/java/org/stellar/sdk/Sep10Challenge.java
@@ -145,7 +145,7 @@ public class Sep10Challenge {
    * @throws InvalidSep10ChallengeException If the SEP-0010 validation fails, the exception will be thrown.
    * @throws IOException                    If read XDR string fails, the exception will be thrown.
    */
-  public static LinkedHashSet<String> verifyChallengeTransactionSigners(String challengeXdr, String serverAccountId, Network network, Set<String> signers) throws InvalidSep10ChallengeException, IOException {
+  public static Set<String> verifyChallengeTransactionSigners(String challengeXdr, String serverAccountId, Network network, Set<String> signers) throws InvalidSep10ChallengeException, IOException {
     if (signers == null || signers.isEmpty()) {
       throw new InvalidSep10ChallengeException("No signers provided.");
     }
@@ -193,7 +193,7 @@ public class Sep10Challenge {
     // are consumed only once on the transaction.
     Set<String> allSigners = new HashSet<String>(clientSigners);
     allSigners.add(serverKeyPair.getAccountId());
-    LinkedHashSet<String> signersFound = verifyTransactionSignatures(transaction, allSigners);
+    Set<String> signersFound = verifyTransactionSignatures(transaction, allSigners);
 
     // Confirm the server is in the list of signers found and remove it.
     boolean serverSignerFound = signersFound.remove(serverKeyPair.getAccountId());
@@ -233,7 +233,7 @@ public class Sep10Challenge {
    * @throws InvalidSep10ChallengeException If the SEP-0010 validation fails, the exception will be thrown.
    * @throws IOException                    If read XDR string fails, the exception will be thrown.
    */
-  public static LinkedHashSet<String> verifyChallengeTransactionThreshold(String challengeXdr, String serverAccountId, Network network, int threshold, Set<Signer> signers) throws InvalidSep10ChallengeException, IOException {
+  public static Set<String> verifyChallengeTransactionThreshold(String challengeXdr, String serverAccountId, Network network, int threshold, Set<Signer> signers) throws InvalidSep10ChallengeException, IOException {
     if (signers == null || signers.isEmpty()) {
       throw new InvalidSep10ChallengeException("No signers provided.");
     }
@@ -243,7 +243,7 @@ public class Sep10Challenge {
       accountIdSignerMap.put(signer.getKey(), signer);
     }
 
-    LinkedHashSet<String> signersFound = verifyChallengeTransactionSigners(challengeXdr, serverAccountId, network, accountIdSignerMap.keySet());
+    Set<String> signersFound = verifyChallengeTransactionSigners(challengeXdr, serverAccountId, network, accountIdSignerMap.keySet());
 
     int weight = 0;
     for (String signer : signersFound) {
@@ -260,7 +260,7 @@ public class Sep10Challenge {
     return signersFound;
   }
 
-  private static LinkedHashSet<String> verifyTransactionSignatures(Transaction transaction, Set<String> signers) throws InvalidSep10ChallengeException {
+  private static Set<String> verifyTransactionSignatures(Transaction transaction, Set<String> signers) throws InvalidSep10ChallengeException {
     List<DecoratedSignature> signatures = transaction.getSignatures();
     if (signatures.isEmpty()) {
       throw new InvalidSep10ChallengeException("Transaction has no signatures.");
@@ -270,7 +270,7 @@ public class Sep10Challenge {
 
     // find and verify signatures
     Set<Integer> signatureUsed = new HashSet<Integer>();
-    LinkedHashSet<String> signersFound = new LinkedHashSet<String>();
+    Set<String> signersFound = new HashSet<String>();
     for (String signer : signers) {
       KeyPair keyPair = KeyPair.fromAccountId(signer);
       int index = -1;
--- a/src/main/java/org/stellar/sdk/Sep10Challenge.java
+++ b/src/main/java/org/stellar/sdk/Sep10Challenge.java
@@ -119,7 +119,13 @@ public class Sep10Challenge {
     }
 
     BaseEncoding base64Encoding = BaseEncoding.base64();
-    byte[] nonce = base64Encoding.decode(new String(manageDataOperation.getValue()));
+    byte[] nonce;
+    try {
+      nonce = base64Encoding.decode(new String(manageDataOperation.getValue()));
+    } catch (IllegalArgumentException e) {
+      throw new InvalidSep10ChallengeException("Failed to decode random nonce provided in ManageData operation.", e);
+    }
+
     if (nonce.length != 48) {
       throw new InvalidSep10ChallengeException("Random nonce before encoding as base64 should be 48 bytes long.");
     }
--- a/src/main/java/org/stellar/sdk/Sep10Challenge.java
+++ b/src/main/java/org/stellar/sdk/Sep10Challenge.java
@@ -240,23 +240,23 @@ public class Sep10Challenge {
       throw new InvalidSep10ChallengeException("No signers provided.");
     }
 
-    Map<String, Signer> accountIdSignerMap = new HashMap<String, Signer>();
+    Map<String, Integer> weightsForSigner = new HashMap<String, Integer>();
     for (Signer signer : signers) {
-      accountIdSignerMap.put(signer.getKey(), signer);
+      weightsForSigner.put(signer.getKey(), signer.getWeight());
     }
 
-    Set<String> signersFound = verifyChallengeTransactionSigners(challengeXdr, serverAccountId, network, accountIdSignerMap.keySet());
+    Set<String> signersFound = verifyChallengeTransactionSigners(challengeXdr, serverAccountId, network, weightsForSigner.keySet());
 
-    int weight = 0;
+    int sum = 0;
     for (String signer : signersFound) {
-      if (!accountIdSignerMap.containsKey(signer)) {
-        continue;
+      Integer weight = weightsForSigner.get(signer);
+      if (weight != null) {
+        sum += weight;
       }
-      weight += accountIdSignerMap.get(signer).getWeight();
     }
 
-    if (weight < threshold) {
-      throw new InvalidSep10ChallengeException(String.format("Signers with weight %d do not meet threshold %d.", weight, threshold));
+    if (sum < threshold) {
+      throw new InvalidSep10ChallengeException(String.format("Signers with weight %d do not meet threshold %d.", sum, threshold));
     }
 
     return signersFound;
--- a/src/main/java/org/stellar/sdk/Sep10Challenge.java
+++ b/src/main/java/org/stellar/sdk/Sep10Challenge.java
@@ -162,12 +162,14 @@ public class Sep10Challenge {
     Set<String> clientSigners = new HashSet<String>();
     for (String signer : signers) {
       // Ignore non-G... account/address signers.
-      Optional<StrKey.VersionByte> versionByteOptional;
-      versionByteOptional = StrKey.decodedVersionByte(signer);
-      if (!versionByteOptional.isPresent()) {
+      StrKey.VersionByte versionByte;
+      try {
+        versionByte = StrKey.decodeVersionByte(signer);
+      } catch (Exception e) {
         continue;
       }
-      if (!StrKey.VersionByte.ACCOUNT_ID.equals(versionByteOptional.get())) {
+
+      if (!StrKey.VersionByte.ACCOUNT_ID.equals(versionByte)) {
         continue;
       }
 
--- a/src/main/java/org/stellar/sdk/Sep10Challenge.java
+++ b/src/main/java/org/stellar/sdk/Sep10Challenge.java
@@ -1,8 +1,14 @@
 package org.stellar.sdk;
 
+import com.google.common.base.Objects;
 import com.google.common.io.BaseEncoding;
+import org.stellar.sdk.xdr.DecoratedSignature;
 
+import java.io.IOException;
 import java.security.SecureRandom;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 
 public class Sep10Challenge {
   /**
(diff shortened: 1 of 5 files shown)


src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +46,366 @@ public static String newChallenge(
...
+   * @throws InvalidSep10ChallengeException If the SEP-0010 validation fails, the exception will be thrown.
+   * @throws IOException                    If read XDR string fails, the exception will be thrown.
+   */
+  public static List<String> verifyChallengeTransactionSigners(String challengeTransaction, String serverAccountId, Network network, List<String> signers) throws InvalidSep10ChallengeException, IOException {
+    if (signers == null || signers.isEmpty()) {
+      throw new InvalidSep10ChallengeException("No signers provided.");
+    }
+
+    // Read the transaction which validates its structure.
+    ChallengeTransaction readableChallengeTransaction = readChallengeTransaction(challengeTransaction, serverAccountId, network);
```
tamirms: I prefer the variable name `parsedChallengeTransaction`

tamirms: or just `parsed`

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +46,366 @@ public static String newChallenge(
...
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Used to store the results produced by {@link Sep10Challenge#readChallengeTransaction(String, String, Network)}.
+   */
+  public static class ChallengeTransaction {
+    private Transaction transaction;
```
tamirms: it seems like `ChallengeTransaction` should be immutable. I would make these fields final and remove their setters

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +46,366 @@ public static String newChallenge(
...
+      return Objects.equal(this.transaction, other.transaction) &&
+              Objects.equal(this.clientAccountId, other.clientAccountId);
+    }
+  }
+
+  /**
+   * Represents signers.
+   */
+  public static class Signer {
+    private String key;
```
tamirms: it seems like `Signer ` should be immutable. I would make these fields final and remove their setters

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +46,366 @@ public static String newChallenge(
...
+      }
+
+      ChallengeTransaction other = (ChallengeTransaction) object;
+      return Objects.equal(this.transaction, other.transaction) &&
+              Objects.equal(this.clientAccountId, other.clientAccountId);
+    }
+  }
+
+  /**
+   * Represents signers.
```
tamirms: Represents a transaction signer.

src/main/java/org/stellar/sdk/Sep10Challenge.java
leighmcculloch: Or `readChallengeTransaction`? The past tense is needed though instead of the `able`.

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +46,366 @@ public static String newChallenge(
...
+   * @param serverAccountId      Account ID for server's account.
+   * @param network              The network to connect to for verifying and retrieving.
+   * @param threshold            The threshold on the client account.
+   * @param signers              The signers of client account.
+   * @return a list of signers that were found is returned, excluding the server account ID.
+   * @throws InvalidSep10ChallengeException If the SEP-0010 validation fails, the exception will be thrown.
+   * @throws IOException                    If read XDR string fails, the exception will be thrown.
+   */
+  public static List<String> verifyChallengeTransactionThreshold(String challengeTransaction, String serverAccountId, Network network, int threshold, List<Signer> signers) throws InvalidSep10ChallengeException, IOException {
+    String ed25519PublicKeySignerType = "ed25519_public_key";
```
tamirms: you can make this a static final member of  `Sep10Challenge`

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +46,366 @@ public static String newChallenge(
...
+   * <p>
+   * It also verifies that transaction is signed by the server.
+   * <p>
+   * It does not verify that the transaction has been signed by the client or
+   * that any signatures other than the servers on the transaction are valid. Use
+   * one of the following functions to completely verify the transaction:
+   * {@link Sep10Challenge#verifyChallengeTransactionSigners(String, String, Network, List)} or
+   * {@link Sep10Challenge#verifyChallengeTransactionThreshold(String, String, Network, int, List)}
+   *
+   * @param challengeTransaction SEP-0010 transaction challenge transaction in base64.
```
tamirms: I prefer the variable name `challengeXdr`

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +46,366 @@ public static String newChallenge(
...
+   * @param serverAccountId      Account ID for server's account.
+   * @param network              The network to connect to for verifying and retrieving.
+   * @return {@link ChallengeTransaction}, the decoded transaction envelope and client account ID contained within.
+   * @throws InvalidSep10ChallengeException If the SEP-0010 validation fails, the exception will be thrown.
+   * @throws IOException                    If read XDR string fails, the exception will be thrown.
+   */
+  public static ChallengeTransaction readChallengeTransaction(String challengeTransaction, String serverAccountId, Network network) throws InvalidSep10ChallengeException, IOException {
```
tamirms: given that this function is taking an xdr representation of a transaction and parsing it into a `ChallengeTransaction` instance, I prefer the name `parseChallengeTransaction()`

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +46,366 @@ public static String newChallenge(
...
+    // Read the transaction which validates its structure.
+    ChallengeTransaction readableChallengeTransaction = readChallengeTransaction(challengeTransaction, serverAccountId, network);
+    Transaction transaction = readableChallengeTransaction.getTransaction();
+
+    // Ensure the server account ID is an address and not a seed.
+    KeyPair serverKeyPair = KeyPair.fromAccountId(serverAccountId);
+
+    // Deduplicate the client signers and ensure the server is not included
+    // anywhere we check or output the list of signers.
+    List<String> clientSigners = new ArrayList<String>();
```
tamirms: It seems like `clientSigners` should be a `Set<String>`

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +46,366 @@ public static String newChallenge(
...
+    }
+
+    if (weight < threshold) {
+      throw new InvalidSep10ChallengeException(String.format("Signers with weight %d do not meet threshold %d.", weight, threshold));
+    }
+
+    return signersFound;
+  }
+
+  private static List<String> verifyTransactionSignatures(Transaction transaction, List<String> signers) throws InvalidSep10ChallengeException {
```
tamirms: if you change the signature of this function to take `Set<String> signers` and return `Set<String>` then you could implement it as:

```java
    return Sets.filter(signers, new Predicate<String>() {
      @Override
      public boolean apply(@NullableDecl String signer) {
        return verifyTransactionSignedByAccountId(transaction, signer);
      }
    });

```

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +46,366 @@ public static String newChallenge(
...
+   * @throws IOException                    If read XDR string fails, the exception will be thrown.
+   */
+  public static List<String> verifyChallengeTransactionThreshold(String challengeTransaction, String serverAccountId, Network network, int threshold, List<Signer> signers) throws InvalidSep10ChallengeException, IOException {
+    String ed25519PublicKeySignerType = "ed25519_public_key";
+
+    if (signers == null || signers.isEmpty()) {
+      throw new InvalidSep10ChallengeException("No signers provided.");
+    }
+
+    List<String> ed25519PublicKeySigners = new ArrayList<String>();
```
tamirms: Seems like this should be a Map<String, Signer>

tamirms:
@overcat thank you for your contribution!

I think if you use Sets and Maps instead of List the code will be simpler. Let me know what you think.

src/main/java/org/stellar/sdk/Sep10Challenge.java
leighmcculloch: I think @overcat was keeping this similar and 1-1 with the Go and Python SDKs that already use the term `read`. Do you see any big issues continuing forward and using read here too?

tamirms: @leighmcculloch ok in that case we can keep the function name as `readChallengeTransaction `

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +46,366 @@ public static String newChallenge(
...
+        continue;
+      }
+      signersFound.add(signer);
+    }
+
+    // Confirm we matched a signature to the server signer.
+    if (!serverSignerFound) {
+      throw new InvalidSep10ChallengeException(String.format("Transaction not signed by server: %s.", serverAccountId));
+    }
+
```
leighmcculloch: I think we need to add a check here similar to the Go SDK that if the signers found after removing the SDK is length 0, then that should be an exception.  https://github.com/stellar/go/blob/a135d590c455964bf1bfdbb33ba266b5a08f9e53/txnbuild/transaction.go#L608-L610

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +46,366 @@ public static String newChallenge(
...
+    List<DecoratedSignature> signatures = transaction.getSignatures();
+    if (signatures.isEmpty()) {
+      throw new InvalidSep10ChallengeException("Transaction has no signatures.");
+    }
+    List<String> signersFound = new ArrayList<String>();
+    for (String signer : signers) {
+      if (verifyTransactionSignedByAccountId(transaction, signer)) {
+        signersFound.add(signer);
+      }
+    }
```
leighmcculloch: This logic verifies each signer has signed the transaction independently, but it would be more defensive to ensure that all signatures are matched to a signer in the list exactly once. Without doing that this code is making the assumption that whatever keys are in use will always work in such a way that no signature could be matched to more than one key and might one day result in two signers being matched to a single signature and a signature being double counted.

@overcat I raised this on the Python SDK also here https://github.com/StellarCN/py-stellar-base/issues/252.

It would be better if we can use the same logic as is used in the Go SDK for signature verification:
https://github.com/stellar/go/blob/a135d590c455964bf1bfdbb33ba266b5a08f9e53/txnbuild/transaction.go#L652-L695

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +46,366 @@ public static String newChallenge(
...
+              Objects.equal(this.clientAccountId, other.clientAccountId);
+    }
+  }
+
+  /**
+   * Represents signers.
+   */
+  public static class Signer {
+    private String key;
+    private String type;
```
leighmcculloch: I think for now we could drop this field and expect a user to only pass in signers of a supported type.

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +46,366 @@ public static String newChallenge(
...
+  public static List<String> verifyChallengeTransactionThreshold(String challengeTransaction, String serverAccountId, Network network, int threshold, List<Signer> signers) throws InvalidSep10ChallengeException, IOException {
+    String ed25519PublicKeySignerType = "ed25519_public_key";
+
+    if (signers == null || signers.isEmpty()) {
+      throw new InvalidSep10ChallengeException("No signers provided.");
+    }
+
+    List<String> ed25519PublicKeySigners = new ArrayList<String>();
+    for (Signer signer : signers) {
+      if (ed25519PublicKeySignerType.equals(signer.getType())) {
```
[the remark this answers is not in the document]
leighmcculloch: FYI I think we can solve this soon, I'd just like us to solve it in a way that we can do consistently, and I don't want it to block this PR.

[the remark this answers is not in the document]
leighmcculloch: Actually, let's solve this now. It occurred to me we can remove the dependency on Horizon by parsing the strkey and ignoring strkeys that are not a [`StrKey.VersionByte.ACCOUNT_ID`](https://github.com/stellar/java-stellar-sdk/blob/bdf9d21e554e82fad466b7dfc39cdd090d13fdb1/src/main/java/org/stellar/sdk/StrKey.java#L14).

I've opened a PR adding this behavior to the Go SDK:
https://github.com/stellar/go/pull/2215

Could you add it to this change in the same way? We'll need to add a function to the `StrKey` class to get the VersionByte from a strkey.

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +46,366 @@ public static String newChallenge(
...
+    }
+
+    if (!verifyTransactionSignedByAccountId(transaction, serverAccountId)) {
+      throw new InvalidSep10ChallengeException(String.format("Transaction not signed by server: %s.", serverAccountId));
+    }
+
+    // verify that transaction sequenceNumber is equal to zero
+    if (transaction.getSequenceNumber() != 0L) {
+      throw new InvalidSep10ChallengeException("The transaction sequence number should be zero.");
+    }
```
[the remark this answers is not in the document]
overcat: I am following the order in the [current proposal](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#token), maybe we should modify it.

src/main/java/org/stellar/sdk/StrKey.java
```diff
@@ -22,6 +22,15 @@
...
             return value;
         }
+
+        public static VersionByte findByValue(byte value) {
+            for (VersionByte versionByte : values()) {
+                if (value == versionByte.value) {
+                    return versionByte;
+                }
+            }
+            throw new IllegalArgumentException("No matching versionByte found.");
```
overcat: Should we throw a custom exception here?

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +44,336 @@ public static String newChallenge(
...
+    // checked in the readChallengeTx to ensure that every signature and signer
+    // are consumed only once on the transaction.
+    Set<String> allSigners = new HashSet<String>(clientSigners);
+    allSigners.add(serverKeyPair.getAccountId());
+    LinkedHashSet<String> allSignersFound = verifyTransactionSignatures(transaction, allSigners);
+
+    // Confirm the server is in the list of signers found and remove it.
+    boolean serverSignerFound = allSignersFound.remove(serverKeyPair.getAccountId());
+    // After removing the server signer we call it clientSignersFound
+    LinkedHashSet<String> clientSignersFound = allSignersFound;
```
overcat: Does this look strange?

src/main/java/org/stellar/sdk/StrKey.java
leighmcculloch: Yeah probably. If we were supporting Java 8+ we could return an Optional but I think we might be supporting older versions of Java. @tamirms do you know?

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +44,336 @@ public static String newChallenge(
...
+   * @param challengeXdr    SEP-0010 transaction challenge transaction in base64.
+   * @param serverAccountId Account ID for server's account.
+   * @param network         The network to connect to for verifying and retrieving.
+   * @param threshold       The threshold on the client account.
+   * @param signers         The signers of client account.
+   * @return a list of signers that were found is returned, excluding the server account ID.
+   * @throws InvalidSep10ChallengeException If the SEP-0010 validation fails, the exception will be thrown.
+   * @throws IOException                    If read XDR string fails, the exception will be thrown.
+   */
+  public static LinkedHashSet<String> verifyChallengeTransactionThreshold(String challengeXdr, String serverAccountId, Network network, int threshold, Set<Signer> signers) throws InvalidSep10ChallengeException, IOException {
```
leighmcculloch: What's the motivation for returning a specific type of Set here? Can we use the generic Set interface?

src/main/java/org/stellar/sdk/Sep10Challenge.java
leighmcculloch: A little. I'd get rid of allSignersFound and just call it signersFound from the start.

src/main/java/org/stellar/sdk/StrKey.java
tamirms: we're limited to using Java 6 because most android devices are stuck on that version. However, you can use com.google.common.base.Optional if you prefer . We use it in other parts of the sdk

src/main/java/org/stellar/sdk/Sep10Challenge.java
overcat: In my opinion, `SignersFound` is in order. This order is the order in which the signers sign. When we return `LinkedHashSet`, the user can explicitly determine that it is in order. 

If I have something wrong please correct me. Thank you.

src/main/java/org/stellar/sdk/StrKey.java
overcat: Thank you, I think it's a good idea to use `com.google.common.base.Optional` when we can't upgrade to Java 8+.

src/main/java/org/stellar/sdk/Sep10Challenge.java
overcat: I'm sorry, I got it wrong, I have fixed it.

overcat: But I think this is worth mentioning, should we return an ordered `signersFound`? This order corresponds to the signatures in the transaction.

src/main/java/org/stellar/sdk/StrKey.java
```diff
@@ -42,6 +52,12 @@ public static AccountID encodeToXDRAccountId(String data) {
         return accountID;
     }
 
+    public static Optional<VersionByte> decodedVersionByte(String data) {
```
leighmcculloch:
```suggestion
    public static Optional<VersionByte> decodeVersionByte(String data) {
```
Can we rename this function for consistency with the other function names?

src/main/java/org/stellar/sdk/StrKey.java
```diff
@@ -42,6 +52,12 @@ public static AccountID encodeToXDRAccountId(String data) {
...
+        byte[] decoded = StrKey.base32Encoding.decode(java.nio.CharBuffer.wrap(data.toCharArray()));
+        byte decodedVersionByte = decoded[0];
+        return VersionByte.findByValue(decodedVersionByte);
```
leighmcculloch: The `Optional` usage in `findByValue` is good, because it is a small utility function and using Optional makes it easier to compose into different situations. Also the name findX returning an absent is easily expected and the two go together well.

For this function, `decodeVersionByte`, I think it is more appropriate to turn the missing result from `findByValue` into a `FormatException` in the same way as the other decode functions do, because given valid input the decode process will always result in a known version byte.

```suggestion
        VersionByte versionByte = VersionByte.findByValue(decodedVersionByte);
        if (!versionByte.isPresent()) {
                throw new FormatException("Version byte is invalid");
        }
        return versionByte.get();
```

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +45,337 @@ public static String newChallenge(
...
+
+  private static Set<String> verifyTransactionSignatures(Transaction transaction, Set<String> signers) throws InvalidSep10ChallengeException {
+    List<DecoratedSignature> signatures = transaction.getSignatures();
+    if (signatures.isEmpty()) {
+      throw new InvalidSep10ChallengeException("Transaction has no signatures.");
+    }
+
+    byte[] txHash = transaction.hash();
+
+    // find and verify signatures
```
tamirms: I think it's possible to simplify this using a Multimap:

```java
    // find and verify signatures
    Set<String> signersFound = new HashSet<String>();
    Multimap<SignatureHint, Signature> signatures = HashMultimap.create();
    for (DecoratedSignature decoratedSignature : transaction.getSignatures()) {
      signatures.put(decoratedSignature.getHint(), decoratedSignature.getSignature());
    }

    for (String signer : signers) {
      KeyPair keyPair = KeyPair.fromAccountId(signer);
      SignatureHint hint = keyPair.getSignatureHint();

      for (Signature signature : signatures.get(hint)) {
        if (keyPair.verify(txHash, signature.getSignature())) {
          signersFound.add(signer);
          // explicitly ensure that a transaction signature cannot be
          // mapped to more than one signer
          signatures.remove(hint, signature);
          break;
        }
      }
    }
```

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +45,337 @@ public static String newChallenge(
...
+   * @return a list of signers that were found is returned, excluding the server account ID.
+   * @throws InvalidSep10ChallengeException If the SEP-0010 validation fails, the exception will be thrown.
+   * @throws IOException                    If read XDR string fails, the exception will be thrown.
+   */
+  public static Set<String> verifyChallengeTransactionThreshold(String challengeXdr, String serverAccountId, Network network, int threshold, Set<Signer> signers) throws InvalidSep10ChallengeException, IOException {
+    if (signers == null || signers.isEmpty()) {
+      throw new InvalidSep10ChallengeException("No signers provided.");
+    }
+
+    Map<String, Signer> accountIdSignerMap = new HashMap<String, Signer>();
```
tamirms: I think it would make the code slightly more readable to use a `Map<String, Integer>` to hold the weights for each signer:

```java
    Map<String, Integer> weightsForSigner = new HashMap<String, Integer>();
    for (Signer signer : signers) {
      weightsForSigner.put(signer.getKey(), signer.getWeight());
    }

    Set<String> signersFound = verifyChallengeTransactionSigners(challengeXdr, serverAccountId, network, weightsForSigner.keySet());

    int sum = 0;
    for (String signer : signersFound) {
      Integer weight = weightsForSigner.get(signer);
      if (weight != null) {
        sum += weight;
      }
    }

    if (sum < threshold) {
      throw new InvalidSep10ChallengeException(String.format("Signers with weight %d do not meet threshold %d.", sum, threshold));
    }
```

src/main/java/org/stellar/sdk/Sep10Challenge.java
overcat: This looks great, but `Multimap` is not a built-in data type. Should we use a third-party data type when it is not necessary?

tamirms: @overcat great point! Multimap comes https://github.com/google/guava . We already have a dependency on guava ( https://github.com/stellar/java-stellar-sdk/blob/master/build.gradle#L45-L46 ) so this change would not require any new dependencies. Note that the `Optional` implementation we're using in the java sdk also comes from guava.

overcat: OK, changed.

leighmcculloch:
>I will defer to @leighmcculloch for the final approval before we can merge

@tamirms Thanks, I've got it in my queue to do a final pass over this PR and I hope to get to it before the end of this week.

@overcat Before I do another pass, I noticed that during my first pass the number of tests seemed smaller than the number of tests we have in the Go SDK. We need to make sure we're covering at least the same test scenarios that are there. I haven't dug into the specifics of the tests here yet, but are there just less because you combined some of the scenarios or are we missing some?

overcat:
Hi, @leighmcculloch, thank you for your reminder, I will add more tests later.

When I checked the test code of Golang, I found that the behavior of [TestReadChallengeTx_invalidNotSignedByServer](https://github.com/stellar/go/blob/master/txnbuild/transaction_test.go#L1226-L1251) was different from what I expected. This should be a bug, maybe [this line](https://github.com/stellar/go/blob/master/txnbuild/transaction.go#L670) should be changed to `if tx.xdrEnvelope.Signatures == nil {`

leighmcculloch:
>When I checked the test code of Golang, I found that the behavior of [TestReadChallengeTx_invalidNotSignedByServer](https://github.com/stellar/go/blob/master/txnbuild/transaction_test.go#L1226-L1251) was different from what I expected. This should be a bug, maybe [this line](https://github.com/stellar/go/blob/master/txnbuild/transaction.go#L670) should be changed to `if tx.xdrEnvelope.Signatures == nil {`

@overcat Good catch, that if statement is guarding on the event someone passes a `Transaction` that has no `xdrEnvelope`, and it is not really testing if the transaction has signatures and the error being returned at that if statement has misleading text. However, I believe the test `TestReadChallengeTx_invalidNotSignedByServer` is still accurate in its expectations. Let me know if you think otherwise.

>thank you for your reminder, I will add more tests later.

Thanks! Please mention me when you've added the tests and I'll take another look, and feel free to mention me if you see any other issues with the tests too. Thanks!

overcat:
Hi @leighmcculloch, I added more tests with reference to the Golang implementation, but I did not add the following tests:
- [TestReadChallengeTx_invalidCorrupted](https://github.com/stellar/java-stellar-sdk/blob/1f9a86c534014b3c922d90c54a7aa3ef7694bb13/src/main/java/org/stellar/sdk/Sep10Challenge.java#L72): I didn't catch the exception in [the code](https://github.com/stellar/java-stellar-sdk/blob/1f9a86c534014b3c922d90c54a7aa3ef7694bb13/src/main/java/org/stellar/sdk/Sep10Challenge.java#L72), should we catch it here? As a newcomer I would love to hear your opinions.
- [TestReadChallengeTx_invalidDataValueCorruptBase64](https://github.com/stellar/go/blob/master/txnbuild/transaction_test.go#L1512): The reason is the same as above.
- [TestVerifyChallengeTxSigners_validServerAndClientSignersIgnoresDuplicateSigner](https://github.com/stellar/go/blob/master/txnbuild/transaction_test.go#L2157): signers is a `Set`, so I don't think we need this.
- [TestVerifyChallengeTxSigners_invalidServerAndClientSignersIgnoresDuplicateSignerInError](TestVerifyChallengeTxSigners_invalidServerAndClientSignersIgnoresDuplicateSignerInError): The reason is the same as above.


BTW, I think [this line](https://github.com/stellar/go/blob/master/txnbuild/transaction_test.go#L2296) should be modified to `err = tx.Sign(serverKP, clientKP2)`.

leighmcculloch:
@overcat Thanks for reporting the unnecessary signer in that test, I'll get that fixed up in stellar/go#2279.

I agree you don't need to add the duplicate signers test this the language protects against that with the use of a Set. 🎉 

Thanks for being so thorough with the addition of more tests. I need a couple days to go over them and to review the questions about the other two tests that you left out. Look out for a follow up from me next week.

Thanks!

src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
```diff
@@ -46,7 +50,1296 @@ public void testChallenge() throws IOException {
...
+            masterClient.getAccountId(),
+            "Stellar Test",
+            timeBounds
+    );
+
+    Transaction transaction = Transaction.fromEnvelopeXdr(challenge, network);
+    transaction.sign(signerClient1);
+
+    try {
+      Sep10Challenge.verifyChallengeTransactionSigners(transaction.toEnvelopeXdrBase64(), server.getAccountId(), network, null);
```
leighmcculloch: ❗️ Could you add a test that is the same as this test but instead of `null` an empty `Set` is passed in?

src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
```diff
@@ -46,7 +50,1296 @@ public void testChallenge() throws IOException {
...
+
+    transaction.sign(client);
+    try {
+      Sep10Challenge.readChallengeTransaction(transaction.toEnvelopeXdrBase64(), server.getAccountId(), network);
+      fail();
+    } catch (InvalidSep10ChallengeException e) {
+      assertEquals(String.format("Transaction not signed by server: %s.", server.getAccountId()), e.getMessage());
+    }
+  }
+
```
leighmcculloch: ❗️ Could we add a test here that tests what happens if the transaction base64 that is passed into the `readChallengeTransaction` function is corrupted?

This is the `TestReadChallengeTx_invalidCorrupted` test in the Go SDK:
https://github.com/stellar/go/blob/f4e38b5/txnbuild/transaction_test.go#L1253-L1281

src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
```diff
@@ -46,7 +50,1296 @@ public void testChallenge() throws IOException {
...
+    String challenge = transaction.toEnvelopeXdrBase64();
+
+    try {
+      Sep10Challenge.readChallengeTransaction(challenge, server.getAccountId(), Network.TESTNET);
+      fail();
+    } catch (InvalidSep10ChallengeException e) {
+      assertEquals("Random nonce encoded as base64 should be 64 bytes long.", e.getMessage());
+    }
+  }
+
```
leighmcculloch: ❗️ Could we add a test here that tests what happens if the data value base64 that is in the challenge transaction is corrupted?

This is the `TestReadChallengeTx_invalidDataValueCorruptBase64` test in the Go SDK:
https://github.com/stellar/go/blob/f4e38b5/txnbuild/transaction_test.go#L1512-L1539

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +48,338 @@ public static String newChallenge(
...
+      ChallengeTransaction other = (ChallengeTransaction) object;
+      return Objects.equal(this.transaction, other.transaction) &&
+              Objects.equal(this.clientAccountId, other.clientAccountId);
+    }
+  }
+
+  /**
+   * Represents a transaction signer.
+   */
+  public static class Signer {
```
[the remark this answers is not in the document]
overcat: This is a great idea, I have added it to my todo list.

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +48,338 @@ public static String newChallenge(
...
+    if (signers == null || signers.isEmpty()) {
+      throw new InvalidSep10ChallengeException("No verifiable signers provided, at least one G... address must be provided.");
+    }
+
+    Map<String, Integer> weightsForSigner = new HashMap<String, Integer>();
+    for (Signer signer : signers) {
+      weightsForSigner.put(signer.getKey(), signer.getWeight());
+    }
+
+    Set<String> signersFound = verifyChallengeTransactionSigners(challengeXdr, serverAccountId, network, weightsForSigner.keySet());
```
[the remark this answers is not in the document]
overcat: Because in my opinion, the weight of the signer is completely useless for `verifyChallengeTransactionSigners`, so I don't think it is necessary to pass it in. Of course, memory consumption is a noteworthy point, thank you for your reminder.

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +48,338 @@ public static String newChallenge(
...
+      }
+
+      if (!(object instanceof Signer)) {
+        return false;
+      }
+
+      Signer other = (Signer) object;
+      return Objects.equal(this.key, other.key) &&
+              Objects.equal(this.weight, other.weight);
+    }
```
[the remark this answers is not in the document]
tamirms: @leighmcculloch yes, I had reviewed the methods before. I also took another look just now and both `hashCode()` and `equals()` look to be correct in `Signer` and `ChallengeTransaction`

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -40,4 +48,344 @@ public static String newChallenge(
...
+    if (manageDataOperation.getValue().length != 64) {
+      throw new InvalidSep10ChallengeException("Random nonce encoded as base64 should be 64 bytes long.");
+    }
+
+    BaseEncoding base64Encoding = BaseEncoding.base64();
+    byte[] nonce;
+    try {
+      nonce = base64Encoding.decode(new String(manageDataOperation.getValue()));
+    } catch (IllegalArgumentException e) {
+      throw new InvalidSep10ChallengeException("Failed to decode random nonce provided in ManageData operation.", e);
```
overcat: @leighmcculloch I added a bit of code here, so it looks better?

src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
overcat: https://github.com/stellar/java-stellar-sdk/blob/dbd69a425b0ac43ce2f95c1e8d2948f2bc095449/src/test/java/org/stellar/sdk/Sep10ChallengeTest.java#L146L172

I think we should throw an exception that fails to read XDR in this [function](https://github.com/stellar/java-stellar-sdk/blob/dbd69a425b/src/main/java/org/stellar/sdk/Transaction.java#L192), so that we can easily catch it elsewhere, just like this: https://github.com/stellar/java-stellar-sdk/pull/264#discussion_r381269520

src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
overcat: https://github.com/stellar/java-stellar-sdk/pull/264#discussion_r381269520 and https://github.com/stellar/java-stellar-sdk/blob/dbd69a425b0ac43ce2f95c1e8d2948f2bc095449/src/test/java/org/stellar/sdk/Sep10ChallengeTest.java#L543L579

src/main/java/org/stellar/sdk/Sep10Challenge.java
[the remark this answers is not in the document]
leighmcculloch: Agreed, but you can pass in the `signers` value which is one of the function parameters and is a `Set<String>`.

src/main/java/org/stellar/sdk/Sep10Challenge.java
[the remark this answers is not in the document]
leighmcculloch: Thanks!

src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
leighmcculloch: Does the function currently not throw an exception trying to decode invalid XDR?

src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
leighmcculloch: Looks good, thanks!

src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
leighmcculloch: As long as the function throws an exception I think you can just catch that here, and we can do any improvement to that in a separate PR.

src/main/java/org/stellar/sdk/Sep10Challenge.java
leighmcculloch: Looks great.

src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
```diff
@@ -143,6 +143,34 @@ public void testReadChallengeTransactionInvalidNotSignedByServer() throws IOExce
...
+    );
+
+    challenge = challenge.replace("A", "B");
+
+    try {
+      Sep10Challenge.readChallengeTransaction(challenge, server.getAccountId(), Network.TESTNET);
+      fail();
+    } catch (RuntimeException ignored) {
+      // TODO: I think we should add an exception signature about the failure to read XDR on Transaction.fromEnvelopeXdr.
+    }
```
leighmcculloch: I think this improvement could happen in a separate PR. In this test i think we just need to assert that an exception occurs if there's nothing about the exception we can assert on.

src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
leighmcculloch: If we can do this: https://github.com/stellar/java-stellar-sdk/pull/264/files/e82649ae8e3b84c159aa7395be8b487177328bdc..dbd69a425b0ac43ce2f95c1e8d2948f2bc095449#r381556847 I think we're good.

leighmcculloch:
@overcat Just one thing left to do, to fill in that `// TODO` in the `testReadChallengeTransactionInvalidCorrupted` test and we can merge it. 🎉

src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
overcat: Actually this is already done here, please see line 168. If `Sep10Challenge.readChallengeTransaction` does not throw an exception, then the test will fail and `fail() function` will throw an AssertionError, so I don't think it needs to be modified here, we can keep this `TODO` as a reminder.

leighmcculloch: Oh! Of cos! Thank you, and sorry I missed that.

overcat:
@leighmcculloch, @tamirms 

Thank you both for your hard work, you have really given me a lot of help, and I have gained a lot here. 

Thank you sincerely. ❤️

PR #265, opened 2020-02-05
Horizon v1.0.0 Compatibility

overcat:
#261 
* feat: add support for `/offers` end-point with query parameters.
* feat: add support for `/accounts` end-point with ?signer and ?asset filters.

### DIFF
--- a/src/main/java/org/stellar/sdk/requests/OffersRequestBuilder.java
+++ b/src/main/java/org/stellar/sdk/requests/OffersRequestBuilder.java
@@ -5,6 +5,8 @@ import okhttp3.HttpUrl;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
+import org.stellar.sdk.Asset;
+import org.stellar.sdk.AssetTypeCreditAlphaNum;
 import org.stellar.sdk.responses.OfferResponse;
 import org.stellar.sdk.responses.Page;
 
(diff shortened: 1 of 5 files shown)


ire-and-curses:
Tamir is out for a week now. @leighmcculloch do you think you could take a look at this?

leighmcculloch:
@ire-and-curses Apologies this fell through my notifications. I don't have much context on what it means to be Horizon v1 compatible and my queue is pretty deep right now so I'm unlikely to get to this PR before @tamirms returns. Let me know if there is particular urgency here I am not aware of.

src/test/java/org/stellar/sdk/requests/AccountsRequestBuilderTest.java
```diff
@@ -18,4 +19,23 @@ public void testAccounts() {
...
+  public void testForSigner() {
+    Server server = new Server("https://horizon-testnet.stellar.org");
+    HttpUrl uri = server.accounts()
+            .forSigner("GDSBCQO34HWPGUGQSP3QBFEXVTSR2PW46UIGTHVWGWJGQKH3AFNHXHXN")
+            .buildUri();
+    assertEquals("https://horizon-testnet.stellar.org/accounts?signer=GDSBCQO34HWPGUGQSP3QBFEXVTSR2PW46UIGTHVWGWJGQKH3AFNHXHXN", uri.toString());
+  }
+
+  @Test
+  public void testForAsset() {
```
tamirms: could you add a test case where the asset is native and another test case where the asset has type AssetTypeCreditAlphaNum12?

src/main/java/org/stellar/sdk/requests/OffersRequestBuilder.java
```diff
@@ -21,16 +23,90 @@ public OffersRequestBuilder(OkHttpClient httpClient, HttpUrl serverURI) {
...
+  }
+
+  /**
+   * Returns all offers buying an asset.
+   *
+   * @param asset The Asset being bought.
+   * @return current {@link OffersRequestBuilder} instance
+   * @see <a href="https://www.stellar.org/developers/horizon/reference/endpoints/offers.html">Offers</a>
+   */
+  public OffersRequestBuilder forBuyingAsset(Asset asset) {
```
tamirms: horizon now suppports `buying` and `selling` query parameters on this endpoint which replace `buying_asset_type`, `buying_asset_code`, etc

https://github.com/stellar/go/pull/2252

let's use the `buying` and `selling` query parameters

src/main/java/org/stellar/sdk/requests/RequestBuilder.java
```diff
@@ -79,25 +79,41 @@ public RequestBuilder order(Order direction) {
...
+   *
+   * @param name  the name of the query parameter
+   * @param asset the asset to be serialized into the query parameter value
+   */
+  public RequestBuilder setAssetParameter(String name, Asset asset) {
+    uriBuilder.setQueryParameter(name, buildAssetParameter(asset));
+    return this;
+  }
+
+  private String buildAssetParameter(Asset asset) {
```
tamirms: I think `encodeAsset()` would be a better name for this function

src/test/java/org/stellar/sdk/requests/OffersRequestBuilderTest.java
```diff
@@ -19,4 +54,46 @@ public void testForAccount() {
...
+
+  @Test
+  public void testForSeller() {
+    Server server = new Server("https://horizon-testnet.stellar.org");
+    HttpUrl uri = server.offers().forSeller("GBRPYHIL2CI3FNQ4BXLFMNDLFJUNPU2HY3ZMFSHONUCEOASW7QC7OX2H").buildUri();
+    assertEquals("https://horizon-testnet.stellar.org/offers?seller=GBRPYHIL2CI3FNQ4BXLFMNDLFJUNPU2HY3ZMFSHONUCEOASW7QC7OX2H", uri.toString());
+  }
+
+  @Test
+  public void testForSellingAsset() {
```
tamirms: could you add a test case for filtering by the native selling asset and another test case for filtering by the native buying asset?

tamirms: also, can you add a test case for filtering by BOTH buying and selling assets since that is supported by the new endpoint?

src/test/java/org/stellar/sdk/requests/AccountsRequestBuilderTest.java
overcat: This parameter should not be a native asset, so I changed the parameter type to `AssetTypeCreditAlphaNum` and added some tests.

overcat:
It is worth mentioning that in horizon 1.0, the `offer_id` in the offers endpoint is a `string` type, but in Java code we will parse it into a `Long` type. Do you think we need to change it? I read the horizon code and it is still of `int64` at the bottom, so I don't think we need to modify it.

PR #266, opened 2020-02-21
Issue 261: Horizon v1.0.0 int to string

cavazosjoe:


src/main/java/org/stellar/sdk/xdr/ManageBuyOfferOp.java
```diff
@@ -53,19 +53,19 @@ public Price getPrice() {
   public void setPrice(Price value) {
     this.price = value;
   }
-  private Int64 offerID;
-  public Int64 getOfferID() {
+  private String64 offerID;
```
overcat: In the latest [XDR definition file](https://github.com/stellar/stellar-core/blob/d99013e685/src/xdr/Stellar-transaction.x#L139L149), the offer id is still `int64`, so I don't think it should be changed to `String64` here.

Do you edit these XDR files by hand? Generally, we will not modify the XDR generated files directly, but use [xdrgen](https://github.com/stellar/xdrgen) to generate these codes.

Please let me know if I am not clear enough.

src/main/java/org/stellar/sdk/xdr/ManageBuyOfferResult.java
```diff
@@ -69,7 +69,7 @@ public int hashCode() {
   }
   @Override
   public boolean equals(Object object) {
-    if (object == null || !(object instanceof ManageBuyOfferResult)) {
+    if (!(object instanceof ManageBuyOfferResult)) {
```
overcat: For the same reason, we should not edit this code.

src/main/java/org/stellar/sdk/xdr/ManageSellOfferOp.java
```diff
@@ -52,19 +52,19 @@ public Price getPrice() {
   public void setPrice(Price value) {
     this.price = value;
   }
-  private Int64 offerID;
-  public Int64 getOfferID() {
+  private String64 offerID;
```
overcat: For the same reason, we should not edit this code.

overcat:
Thank you for your contribution, but this PR is too old and outdated, so I have decided to close it.

PR #268, opened 2020-02-25
0.14.0 Release

tamirms:
* Update challenge transaction helpers for SEP-10 v1.3.0 (https://github.com/stellar/java-stellar-sdk/issues/263).
* Add support for /accounts end-point with ?signer and ?asset filters (https://github.com/stellar/java-stellar-sdk/issues/261).
* Add support for /offers end-point with query parameters (https://github.com/stellar/java-stellar-sdk/issues/261).
* Regenerate the XDR definitions to include MetaV2 support (https://github.com/stellar/java-stellar-sdk/issues/261).


abuiles:
@tamirms I think I forgot to list this, but can you include support for `/offers/:offer_id` too?

build.gradle
```diff
@@ -16,7 +16,7 @@ apply plugin: 'com.github.ben-manes.versions' // gradle dependencyUpdates -Drevi
 apply plugin: 'project-report' // gradle htmlDependencyReport
 
 sourceCompatibility = 1.6
-version = '0.13.0'
+version = '0.14.0'
```
abuiles: @tamirms would it make sense to make this 1.0.0 too?

tamirms: I'm not sure but I don't think the sdk versions have to match horizon's version

abuiles: they don't  -- I was just suggesting also to add extra emphasis that there have been breaking changes through the latest release -- but I'm okay with this :)

abuiles: For what is worth none of the sdks match horizon's version  :)

tamirms: I think semver allows breaking changes without incrementing the major release if the version is less than 1.0.0

PR #269, opened 2020-03-13
Upgrades OkHttp to latest (4.4.1)

Synesso:


ire-and-curses:
May be relevant to #271

PR #270, opened 2020-03-16
Optimize the code in `TimeBounds`.

overcat:
Please check the commit log.

### DIFF
--- a/src/main/java/org/stellar/sdk/TimeBounds.java
+++ b/src/main/java/org/stellar/sdk/TimeBounds.java
@@ -17,7 +17,15 @@ final public class TimeBounds {
 	 * @param maxTime 64bit Unix timestamp
 	 */
 	public TimeBounds(long minTime, long maxTime) {
-		if (maxTime > 0 && minTime >= maxTime) {
+		if (minTime < 0) {
+			throw new IllegalArgumentException("minTime cannot be negative");
+		}
+
+		if (maxTime < 0) {
+			throw new IllegalArgumentException("maxTime cannot be negative");
+		}
+
+		if (maxTime != 0 && minTime > maxTime) {
 			throw new IllegalArgumentException("minTime must be >= maxTime");
 		}
 		
--- a/src/main/java/org/stellar/sdk/TimeBounds.java
+++ b/src/main/java/org/stellar/sdk/TimeBounds.java
@@ -11,7 +11,7 @@ import org.stellar.sdk.xdr.Uint64;
 final public class TimeBounds {
 	final private long mMinTime;
 	final private long mMaxTime;
-	
+
 	/**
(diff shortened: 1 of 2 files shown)


src/main/java/org/stellar/sdk/TimeBounds.java
```diff
@@ -11,24 +11,44 @@
...
 	}
-	
+
+	/**
+	 * A factory method that sets maxTime to the specified second from now.
+	 *
+	 * @param timeout Timeout in seconds.
+	 * @return TimeBounds
+	 */
+	public static TimeBounds afterNow(long timeout) {
```
overcat: [Duration](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html) was added in Java 8, so we cannot use it for compatibility.

overcat:
@tamirms Can you check this PR? It contains bug fixes and maybe we can include it in the new release.

src/main/java/org/stellar/sdk/TimeBounds.java
```diff
@@ -61,12 +81,16 @@ public static TimeBounds fromXdr(org.stellar.sdk.xdr.TimeBounds timeBounds) {
...
-
+		if (this == o) {
+			return true;
+		}
+		if (o == null || getClass() != o.getClass()) {
+			return false;
+		}
 		TimeBounds that = (TimeBounds) o;
-
-		if (mMinTime != that.mMinTime) return false;
```
tamirms: I think it's easier to read
```java
return mMinTime == that.mMinTime && mMaxTime == that.mMaxTime;
```

instead of:
```java
		if (mMinTime != that.mMinTime) {
			return false;
		}
		return mMaxTime == that.mMaxTime;
```

tamirms:
@overcat the pr looks good. once you address https://github.com/stellar/java-stellar-sdk/pull/270#discussion_r400137770 I can merge it

src/main/java/org/stellar/sdk/TimeBounds.java
overcat: Thanks, done.

overcat: sorry, wait for me to fix the commit log

overcat:
> @overcat the pr looks good. once you address [#270 (comment)](https://github.com/stellar/java-stellar-sdk/pull/270#discussion_r400137770) I can merge it

Already done.

src/main/java/org/stellar/sdk/TimeBounds.java
tamirms: actually, I think `expiresAfter(long seconds)` would be a better method name

overcat: Thanks, done.

PR #273, opened 2020-03-27
feat: add support for SEP-0029.

### LINKED ISSUE #272: Add support for SEP0029
Add support for [SEP 29](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0029.md). This is important to provide a simple way to protect users from forgetting memos when submitting payments to exchanges.

- [ ] Add `server.checkMemoRequired(tx)` with support for [SEP0029](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0029.md).
    If any of the operations included in `tx` are of type `payment`, `pathPaymentStrictReceive`, `pathPaymentStrictSend`, or `mergeAccount`, then the SDK will load the destination account from Horizon and check if `config.memo_required` is set to `1` as defined in [SEP0029](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0029.md).

    - While you could do the check in parallel, we decided to do a sequential check to avoid getting rate limited by SDF's public horizon which is used by default by many clients (read more [here](https://github.com/stellar/js-stellar-sdk/pull/516#discussion_r398168886)).

- [ ] Change `server.SubmitTransaction` to always check if memo is required.
-  [ ] Allow the the check to be bypassed by setting `skipMemoRequiredCheck` to `true`, but default to requiring the check:

	```
	server.SubmitTransaction(tx, {skipMemoRequiredCheck: true})
	```

- [ ] The check should  be skipped automatically if the transaction includes a memo.

You can use the following implementations as a reference:

 - `JS SDK`: https://github.com/stellar/js-stellar-sdk/pull/516
 - `Go SDK`: https://github.com/stellar/go/pull/2396
 
 
Refs: https://github.com/stellar/stellar-protocol/issues/575

overcat:
resolve #272

### DIFF
--- a/src/main/java/org/stellar/sdk/Server.java
+++ b/src/main/java/org/stellar/sdk/Server.java
@@ -10,6 +10,8 @@ import org.stellar.sdk.responses.*;
 import java.io.Closeable;
 import java.io.IOException;
 import java.net.SocketTimeoutException;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
@@ -32,6 +34,17 @@ public class Server implements Closeable {
      * after internal txsub timeout.
      */
     private static final int HORIZON_SUBMIT_TIMEOUT = 60;
+
+    /**
+     * ACCOUNT_REQUIRES_MEMO_VALUE is the base64 encoding of "1".
+     * SEP 29 uses this value to define transaction memo requirements for incoming payments.
+     */
+    public static final String ACCOUNT_REQUIRES_MEMO_VALUE = "MQ==";
+
+    /**
+     * ACCOUNT_REQUIRES_MEMO_KEY is the data name described in SEP 29.
+     */
+    public static final String ACCOUNT_REQUIRES_MEMO_KEY = "config.memo_required";
 
     public Server(String uri) {
         this(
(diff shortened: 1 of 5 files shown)


src/main/java/org/stellar/sdk/Server.java
```diff
@@ -264,6 +272,65 @@ public SubmitTransactionResponse submitTransaction(Transaction transaction) thro
...
+    public void checkMemoRequired(Transaction transaction) throws IOException, AccountRequiresMemoException {
+        if (!(transaction.getMemo() == null || transaction.getMemo().equals(Memo.none()))) {
+            return;
+        }
+        Set<String> destinations = new HashSet<String>();
+        Operation[] operations = transaction.getOperations();
+        for (int i = 0; i < operations.length; i++) {
+            String destination;
+            Operation operation = operations[i];
+            if (operation instanceof PaymentOperation) {
```
overcat: Is there a better way to do this under Java 6?

src/main/java/org/stellar/sdk/Server.java
```diff
@@ -264,6 +272,65 @@ public SubmitTransactionResponse submitTransaction(Transaction transaction) thro
...
+            } else if (operation instanceof AccountMergeOperation) {
+                destination = ((AccountMergeOperation) operation).getDestination();
+            } else {
+                continue;
+            }
+            if (destinations.contains(destination)) {
+                continue;
+            }
+            destinations.add(destination);
+            AccountResponse.Data data = this.accounts().account(destination).getData();
```
overcat: Instead of doing more checks here, I throw the exception to the caller.

src/main/java/org/stellar/sdk/Server.java
tamirms: I don't think so

PR #276, opened 2020-04-22
Implement partial protocol 13 support

tamirms:
* Add support for new trustline flags
* Update strkey to handle muxed accounts
* Update Transaction class, specifically the conversion code to and from XDR

This PR still does not handle fee bump transactions and updates to SEP 10 validation. I will add support for both in the next PR.


src/main/java/org/stellar/sdk/AllowTrustOperation.java
```diff
@@ -94,7 +105,18 @@ public boolean getAuthorize() {
         default:
           throw new RuntimeException("Unknown asset code");
       }
-      authorize = op.getAuthorize();
+
+      int flag = op.getAuthorize().getUint32().intValue();
+      if (flag == TrustLineFlags.AUTHORIZED_FLAG.getValue()) {
```
bartekn, line 110: We should probably error on AUTHORIZED_FLAG & AUTHORIZED_TO_MAINTAIN_LIABILITIES_FLAG != 0.

src/main/java/org/stellar/sdk/StrKey.java
```diff
@@ -41,6 +43,32 @@ public static String encodeStellarAccountId(byte[] data) {
...
+            return encodeStellarAccountId(account.getEd25519().getUint256());
+        } else if (account.getDiscriminant().equals(CryptoKeyType.KEY_TYPE_MUXED_ED25519)) {
+            ByteArrayOutputStream byteStream = new ByteArrayOutputStream();
+
+            try {
+                account.getMed25519().getId().encode(new XdrDataOutputStream(byteStream));
+                byteStream.write(account.getMed25519().getEd25519().getUint256());
+            } catch (IOException e) {
+                throw new IllegalArgumentException("invalid muxed account", e);
+            }
```
bartekn, line 63: Why not use `MuxedAccountMed25519.encode` here?

src/main/java/org/stellar/sdk/StrKey.java
```diff
@@ -52,6 +80,37 @@ public static AccountID encodeToXDRAccountId(String data) {
...
+            muxedAccount.setDiscriminant(CryptoKeyType.KEY_TYPE_MUXED_ED25519);
+            MuxedAccount.MuxedAccountMed25519 m = new MuxedAccount.MuxedAccountMed25519();
+            try {
+                m.setId(Uint64.decode(
+                    new XdrDataInputStream(new ByteArrayInputStream(decoded, 0, 8))
+                ));
+                m.setEd25519(Uint256.decode(
+                    new XdrDataInputStream(new ByteArrayInputStream(decoded, 8, decoded.length - 8))
+                ));
+            } catch (IOException e) {
```
bartekn, line 104: Why not use `MuxedAccountMed25519.decode` here?

src/main/java/org/stellar/sdk/StrKey.java
```diff
@@ -102,6 +166,10 @@ public static String encodeSha256Hash(byte[] data) {
             outputStream.write(checksum);
             byte unencoded[] = outputStream.toByteArray();
 
+            if (VersionByte.SEED != versionByte) {
+                return StrKey.base32Encoding.encode(unencoded).toCharArray();
+            }
```
bartekn, line 171: We can probably remove `if (VersionByte.SEED == versionByte) {` line below.

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -154,28 +175,28 @@ public int getFee() {
   /**
    * Generates Transaction XDR object.
    */
-  public org.stellar.sdk.xdr.Transaction toXdr() {
+  public TransactionV0 toXdr() {
```
bartekn, line 178: Isn't `xdr.Transaction` a default now?

```suggestion
  public Transaction toXdr() {
```

tamirms: until protocol 13 is enabled the sdks should still be generating v0 transactions. We will do another release after protocol 13 is enabled so that the sdks start generating v1 transactions. But, this method should be private the SDK users can get the xdr envelope. they don't need a public method to get the transaction from the envelope. I will change this method to private

src/main/java/org/stellar/sdk/StrKey.java
```diff
@@ -130,6 +196,10 @@ public static String encodeSha256Hash(byte[] data) {
     }
 
     protected static byte[] decodeCheck(VersionByte versionByte, char[] encoded) {
+        if (encoded.length == 0) {
+            throw new IllegalArgumentException("Encoded char array cannot be empty.");
+        }
+
         byte[] bytes = new byte[encoded.length];
         for (int i = 0; i < encoded.length; i++) {
             if (encoded[i] > 127) {
```
2opremio, line 205: Not relevant to this PR, but why is this necesary? `base32Encoding.decode` should be taking care of this, shouldn't it?

src/main/java/org/stellar/sdk/StrKey.java
```diff
@@ -138,6 +208,19 @@ public static String encodeSha256Hash(byte[] data) {
             bytes[i] = (byte) encoded[i];
         }
 
+        int leftoverBits = (bytes.length * 5) % 8;
+        if (leftoverBits > 0) {
```
2opremio, line 212: If `leftoverBits` >= 5 this check would be wrong, because you would be checking a full leftover char, which shouldn't had been there at all.

2opremio: e.g. `MHA` would pass the check, but it should fail because, there are (3*5)%8  = 7 leftover bits, so:

1. `A` shouldn't be there (because it provides 5 unused bits)
2. That leaves 2 extra unused bits in `H`. `H` in base 32 is `00111` , which isn't a proper canonical representation, it should be `00100` (`E`), because the last two bits should be 0.

2opremio: Note also, that, while all these checks are needed in Golang (because the decoder in `"encoding/base32"` swallows non-canonical base32 strings) the Java library maybe be able to do detect this without extra code.

2opremio: `guava` doesn't seem to provide that functionality. There may be other libraries that do, though

src/main/java/org/stellar/sdk/StrKey.java
2opremio: Maybe this is needed due to `guava` not handling it properly ...

src/main/java/org/stellar/sdk/Transaction.java
wumingmo: Our previous code using stellar api is broken now. I was told to use the latest version. One of place broken is that we were getting transaction from transaction envelope. Obviouly, it is broken now. Could you please let me know what is the alternative way to get the transaction from the envelope with current version? Are we suppose to keep compatability in version upgrading?

tamirms: @wumingmo can you please create an issue and provide some example code which demonstrates the broken behavior? thanks!

wumingmo: Thanks, Tamirms, for quick reponse, I created a pull request before, https://github.com/stellar/java-stellar-sdk/issues/284, but was closed and told to use the latest version. Of course, after I use the 0.16.0, I got the compilation error. If possible, could you please let me know how to extract the transaction from the envelope with 0.16.0? Thanks

PR #282, opened 2020-05-26
Rollback SEP23 (Muxed Account strkeys) support

tamirms:
Fix https://github.com/stellar/java-stellar-sdk/issues/277#issuecomment-632876264

### DIFF
--- a/src/main/java/org/stellar/sdk/StrKey.java
+++ b/src/main/java/org/stellar/sdk/StrKey.java
@@ -10,11 +10,9 @@ import java.util.Arrays;
 class StrKey {
 
     public static final int ACCOUNT_ID_ADDRESS_LENGTH = 56;
-    public static final int MUXED_ACCOUNT_ADDRESS_LENGTH = 69;
 
     public enum VersionByte {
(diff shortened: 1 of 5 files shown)


src/main/java/org/stellar/sdk/StrKey.java
```diff
@@ -53,16 +51,7 @@ public static String encodeStellarMuxedAccount(MuxedAccount account) {
...
-
-            try {
-                account.getMed25519().encode(new XdrDataOutputStream(byteStream));
-            } catch (IOException e) {
-                throw new IllegalArgumentException("invalid muxed account", e);
-            }
-
-            char[] encoded = encodeCheck(VersionByte.MUXED_ACCOUNT, byteStream.toByteArray());
-            return String.valueOf(encoded);
+            return encodeStellarAccountId(account.getMed25519().getEd25519().getUint256());
```
2opremio, line 54: Instead of modifying `encodeStellarMuxedAccount` to internally transform to an `AccountId` I would add a method to explicitly transform Muxed Accounts into account IDs. This is more explicit and there is no magic involved.

Issue #292, opened 2020-09-20
Some envelopes with buy/sell offers decoded incorrectly


efritze:
###  What version are you using?

0.18.0

### What did you do?

```java

@Test
    public void offerWithPriceFractionsNotInLowestTerms() throws IOException {

        //numerator is 10000000
        //denominator is 50
        //decode transaction should maintain same values without attempting to reduce fraction to its lowest terms.
        String envelope = "AAAAAgAAAADyfFnjmoCWKyzSBOiKQzP3K8l5z5/uF+pQ2ruXEh9+qAAAAGQADM4dAAAAAQAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAABAAAAAK0KBx2qoQ+o9JNLweoxVw6UYcRUzyKQH/ExRjJ63HEBAAAAAwAAAAFCVEMAAAAAAC8/6Fx60RqPg0dQCYSIhede3rzDSAWgvrFAsTn0LiI8AAAAAAAAAAAAAAAzAJiWgAAAADIAAAAAAAAAAAAAAAAAAAAA";

        AbstractTransaction txn = org.stellar.sdk.Transaction.fromEnvelopeXdr(envelope, Network.TESTNET);

        String returnedEnvelope = txn.toEnvelopeXdrBase64();

        assertEquals(envelope, returnedEnvelope);

    }

```

### What did you expect to see?

Test passed. Decoding and encoding the envelope XDR should reproduce the same envelope XDR.

### What did you see instead?

Test failed.  Decoding and encoding the envelope XDR reproduced a different envelope XDR.

The original envelope had price n = 10000000 and d = 50. The regenerated envelope reduced the price fractions to its lowest terms with n = 200000 and d = 1. Though correct, this generates an XDR which is different from the original XDR.

Prices should be represented internally as fractions, not strings.

efritze:
I submitted a pull request which fixes the issue.

https://github.com/stellar/java-stellar-sdk/pull/293

Issue #183, opened 2019-02-27
Exception thrown from ManageOfferOperation.fromXDR for certain offers


silent-grass:
Hi, Stellar Team

**Version: 0.6.0**
I see the related issue, but nothing was fixed: stellar/java-stellar-sdk: Pull Request 97

Steps to reproduce:
Just invoke: ```org.stellar.sdk.Transaction.fromEnvelopeXdr(AAAAAButy5zasS3DLZ5uFpZHL25aiHUfKRwdv1+3Wp12Ce7XAAAAZAEyGwYAAAAOAAAAAAAAAAAAAAABAAAAAQAAAAAbrcuc2rEtwy2ebhaWRy9uWoh1HykcHb9ft1qddgnu1wAAAAMAAAAAAAAAAUtJTgAAAAAARkrT28ebM6YQyhVZi1ttlwq/dk6ijTpyTNuHIMgUp+EAAAAAAAARPSfDKZ0AAv7oAAAAAAAAAAAAAAAAAAAAAXYJ7tcAAABAbE8rEoFt0Hcv41iwVCl74C1Hyr+Lj8ZyaYn7zTJhezClbc+pTW1KgYFIZOJiGVth2xFnBT1pMXuQkVdTlB3FCw==)```

Result:
```**ArithmeticException: "Non-terminating decimal expansion; no exact representable decimal result" **```

Here
 ```
case MANAGE_OFFER:
        operation = new ManageOfferOperation.Builder(body.getManageOfferOp()).build();
        break;
```

Thanks.

silent-grass:
@nullstyle It can happen for MANAGE_OFFER and CREATE_PASSIVE_OFFER operations.
Point is here: ManageOfferOperation.Builder/CreatePassiveOfferOperation.Builder:
**price = new BigDecimal(n).divide(new BigDecimal(d)).toString()**

Thanks.

nullstyle:
@silent-grass thanks for the heads up.  I should have a fix in place soon, I just need to rework some things and put a regression test in place

PR #293, opened 2020-09-20
Fix offers with price fractions not lowest terms

efritze:
Fixes: https://github.com/stellar/java-stellar-sdk/issues/292

Fixes: https://github.com/stellar/java-stellar-sdk/issues/183

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 As this project is pre 1.0, breaking changes may happen for minor version bumps. A breaking change will get clearly notified in this log.
 
+## 0.18.0
+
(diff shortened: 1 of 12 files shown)


efritze:
The previous version of this pull request https://github.com/stellar/java-stellar-sdk/pull/97 was not approved because it changed the required type of buy and sell offers from String to Price, _breaking_ backward compatibility.

I believe it is trivial to use Price.fromString() for use cases which require string prices. It is more important the SDK produces  consistent results. An envelope XDR decoded and re-encoded should reproduce the same envelope XDR.

overcat:
Hi @efritze, thanks for your contribution, would you like to update this PR? I am willing to assist you in merging this PR.

overcat:
Closing in favor of #554

PR #300, opened 2020-10-16
feat: update challenge transaction helpers for SEP-0010 v2.1.0.

### LINKED ISSUE #298: Implement SEP-10 2.1.0 Support
## TL;DR

The SEP-10 protocol has been updated to relax the requirements of the single Manage Data operation included in SEP-10 challenge transactions, and **should be implemented by October 21st**. On this date the SDF will announce the SDK updates to Anchors and Wallets, urging them to update as soon as possible.

## Background 

To provide context, the [SEP-10 2.0 changes](https://github.com/stellar/stellar-protocol/pull/708) replaced SEP-10 challenges' Manage Data operation `anchorName` key with a `home_domain` key. However, we quickly realized there was confusion and misunderstanding about the value of the `home_domain` parameter introduced in most SDKs earlier. Specifically, some users of the SDK believe the `home_domain` value is the home domain used to request the SEP-10 challenge, while others believe it is the home domain that requires the JWT authentication token.

This is problematic because SDK's implemented equality checks on the `home_domain` value of the Manage Data operation, so mismatched domains would result in client-server (wallet-anchor) incompatibility.

## SEP-10 2.1.0 Changes

The following changes have been made to the SEP-10 protocol:

- Clients are no longer required to validate the `home_domain` value in a SEP-10 challenge's first Manage Data operation
- SEP-10 challenge transactions may now have additional Manage Data operations as long as the source account of the operation is the SEP-10 server's `SIGNING_KEY`
    - Note: additional Manage Data operations will be added in future changes, so any added Manage Data operations added by the issuer of the challenge will need to come *after* the Manage Data operations defined in this document
    
## References

[SEP-10 2.1.0 Changes](https://github.com/stellar/stellar-protocol/pull/708)
[Go SDK SEP-10 2.1.0 Support](https://github.com/stellar/go/pull/3108)
[JavaScript SDK SEP-10 2.1.0 Support](https://github.com/stellar/js-stellar-sdk/pull/580)

## Looking Forward 

Soon after SEP-10 2.1.0 is implmenented and announced, the SDF will ask SDK's to upgrade to SEP-10 3.0, but no action needs to be taken for this yet.

overcat:
resolve #298

### DIFF
--- a/src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
+++ b/src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
@@ -3,6 +3,7 @@ package org.stellar.sdk;
 import com.google.common.collect.ImmutableList;
 import com.google.common.io.BaseEncoding;
 import org.junit.Test;
+import org.stellar.sdk.xdr.BumpSequenceOp;
 import org.stellar.sdk.xdr.EnvelopeType;
 import org.stellar.sdk.xdr.TransactionEnvelope;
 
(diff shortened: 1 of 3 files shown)


src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
```diff
@@ -703,12 +740,35 @@ public void testReadChallengeTransactionInvalidDataValueWrongEncodedLength() thr
...
+
+    Transaction transaction = Sep10Challenge.newChallenge(
+        server,
+        network,
+        client.getAccountId(),
+        domainName,
+        timeBounds
+    );
+
+    Sep10Challenge.ChallengeTransaction challengeTransaction = Sep10Challenge.readChallengeTransaction(transaction.toEnvelopeXdrBase64(), server.getAccountId(), Network.TESTNET, null);
```
leighmcculloch, line 766: Could you add a second test that is similar to this but passes in a value for home domain that is invalid? It's important that we test that if  the home domain is passed it is still not verified.

leighmcculloch (approved):

leighmcculloch:
Thanks @overcat for contributing this change!

src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
```diff
@@ -703,12 +740,35 @@ public void testReadChallengeTransactionInvalidDataValueWrongEncodedLength() thr
...
+    assertEquals(new Sep10Challenge.ChallengeTransaction(transaction, client.getAccountId()), challengeTransaction);
```
JakeUrban, line 767: Why do we need a `ChallengeTransaction` class? I'm not saying we should remove it but I was originally confused on the difference between `Transaction` and `ChallengeTransaction`, since that distinction doesn't exist in the other SDK's. 

It appears to be a collection of helper functions common to challenge transaction as opposed to normal transactions.

JakeUrban (approved):
Looks great, just wanted to clarify if there was more to the `ChallengeTransaction` vs. `Transaction` functionality than I'm seeing.

src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
overcat: Maybe the class name is not so appropriate that makes you confused.

We need it because Java does not support multiple return values.

src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
overcat: Thanks, I have added the tests.

leighmcculloch (approved):
Thanks @overcat !

PR #314, opened 2021-01-07
Regenerate XDR types using latest version of xdrgen

tamirms:
The XDR classes have been regenerated using the latest version of xdrgen which provides the following two features:

stellar/xdrgen#56 Builder static inner classes to unions and structs
stellar/xdrgen#57 Constructors for typedefs


ire-and-curses (approved):
I didn't read all 149 generated files, but the first few look reasonable to me

PR #323, opened 2021-03-09
Update XDR definitions for CAP 35

tamirms:
As part of the work for https://github.com/stellar/java-stellar-sdk/issues/322 , I have updated the XDR definitions in the Java sdk to match those found in the go monorepo. I then ran xdrgen to generate the Java XDR classes. The only adjustment I had to do after generating the new XDR classes was to update src/main/java/org/stellar/sdk/AllowTrustOperation.java to handle the new representation of the asset field in the Allow Trust operation.

In an upcoming PR I will add support for the new operations introduced in CAP 35.


2opremio:
Don't forget the changes at https://github.com/stellar/go/pull/3458

tamirms:
@2opremio thanks, it should be fixed now

2opremio (approved):

PR #325, opened 2021-03-16
Add support for CAP 35 operations when building transactions

tamirms:
This PR adds support for building transactions using the operations introduced in CAP 35. Once this PR is merged then we can close https://github.com/stellar/java-stellar-sdk/issues/322

### DIFF
--- a/src/main/java/org/stellar/sdk/SetTrustlineFlagsOperation.java
+++ b/src/main/java/org/stellar/sdk/SetTrustlineFlagsOperation.java
@@ -1,0 +1,171 @@
+package org.stellar.sdk;
+
+import com.google.common.base.Objects;
+import org.stellar.sdk.xdr.*;
+
+import java.util.EnumSet;
+
+/**
+ *
+ * Represents a Set Trustline Flags operation.
+ * @see <a href="https://www.stellar.org/developers/learn/concepts/list-of-operations.html" target="_blank">List of Operations</a>
+ */
+public class SetTrustlineFlagsOperation extends Operation {
+  private final String trustor;
+  private final AssetTypeCreditAlphaNum asset;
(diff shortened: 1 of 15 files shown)


src/main/java/org/stellar/sdk/SetTrustlineFlagsOperation.java
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  @Override
+  org.stellar.sdk.xdr.Operation.OperationBody toOperationBody() {
+    SetTrustLineFlagsOp op = new SetTrustLineFlagsOp();
+
+    op.setTrustor(StrKey.encodeToXDRAccountId(this.trustor));
+    op.setAsset(asset.toXdr());
+    op.setClearFlags(bitwiseOr(clearFlags));
+    op.setSetFlags(bitwiseOr(setFlags));
```
bartekn, line 71: I think in the [newest version of XDR](https://github.com/stellar/go/blob/b21259c79556a2bf2515e612e879831e1f63fc4d/xdr/Stellar-transaction.x#L409) these fields are nullable which means this code will have to be changed. Are you planning to open a new PR or change it now?

bartekn (approved):

src/main/java/org/stellar/sdk/SetTrustlineFlagsOperation.java
tamirms: actually in the newest version of the xdr the fields are not pointers anymore, see: https://github.com/stellar/go/pull/3458

bartekn: OK, I wasn't aware of this.

2opremio (approved):

PR #336, opened 2021-03-22
util: replaced to/from hex method implementations.

ardevd:
The DataTypeConverter class is supplied with JAXB library. This is part
of the standard library in Java 7.


ardevd:
Apparently not included in the compiler used here after all. Disregard.

PR #348, opened 2021-06-08
Add OPT-IN support for SEP0023

tamirms:
Add opt-in support for SEP 23. SEP23 adds a strkey representation (M-strkeys) for MuxedAccounts.

Fixes https://github.com/stellar/java-stellar-sdk/issues/343

### DIFF
--- a/src/main/java/org/stellar/sdk/AccountConverter.java
+++ b/src/main/java/org/stellar/sdk/AccountConverter.java
@@ -9,15 +9,27 @@ public class AccountConverter {
     this.enableMuxed = enabled;
   }
 
+  /**
+   * Returns an AccountConverter which supports muxed accounts.
+   **/
   public static AccountConverter enableMuxed() {
     return new AccountConverter(true);
   }
 
+  /**
+   * Returns an AccountConverter which does not support muxed accounts. When trying to encode or decode
+   * a muxed account the AccountConverter will first convert the muxed account into an ED25519 account id.
+   **/
   public static AccountConverter disableMuxed() {
     return new AccountConverter(false);
   }
 
 
+  /**
+   * Encodes an account string into its XDR MuxedAccount representation.
+   *
+   * @param account the string representation of an account
+   */
   public MuxedAccount encode(String account) {
     MuxedAccount muxed = StrKey.encodeToXDRMuxedAccount(account);
 
@@ -31,6 +43,11 @@ public class AccountConverter {
     return muxed;
   }
 
+  /**
+   * Decodes an XDR MuxedAccount into its string representation.
+   *
+   * @param account the XDR MuxedAccount representation of an account
+   */
   public String decode(MuxedAccount account) {
     if (this.enableMuxed || account.getDiscriminant().equals(CryptoKeyType.KEY_TYPE_ED25519)) {
       return StrKey.encodeStellarMuxedAccount(account);
--- a/src/main/java/org/stellar/sdk/AccountConverter.java
+++ b/src/main/java/org/stellar/sdk/AccountConverter.java
@@ -1,0 +1,41 @@
+package org.stellar.sdk;
+
+import org.stellar.sdk.xdr.CryptoKeyType;
+import org.stellar.sdk.xdr.MuxedAccount;
+
+public class AccountConverter {
+  private final boolean enableMuxed;
+  private AccountConverter(boolean enabled) {
+    this.enableMuxed = enabled;
+  }
+
+  public static AccountConverter enableMuxed() {
+    return new AccountConverter(true);
+  }
+
+  public static AccountConverter disableMuxed() {
+    return new AccountConverter(false);
+  }
+
+
+  public MuxedAccount encode(String account) {
+    MuxedAccount muxed = StrKey.encodeToXDRMuxedAccount(account);
+
+    if (this.enableMuxed || muxed.getDiscriminant().equals(CryptoKeyType.KEY_TYPE_ED25519)) {
+      return muxed;
+    }
+
+    MuxedAccount.MuxedAccountMed25519 med25519 = muxed.getMed25519();
+    muxed.setDiscriminant(CryptoKeyType.KEY_TYPE_ED25519);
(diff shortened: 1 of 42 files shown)


src/main/java/org/stellar/sdk/AccountConverter.java
```diff
@@ -0,0 +1,41 @@
...
+  public static AccountConverter enableMuxed() {
+    return new AccountConverter(true);
+  }
+
+  public static AccountConverter disableMuxed() {
+    return new AccountConverter(false);
+  }
+
+
+  public MuxedAccount encode(String account) {
```
Shaptic, line 21: This converts an input `account` string in both G and M-form into a `MuxedAccount` with only the G-account representation, right? (that is, forcing the Ed25519 discriminant). A docstring here would be :100:

src/main/java/org/stellar/sdk/AllowTrustOperation.java
```diff
@@ -47,7 +47,7 @@ public boolean getAuthorize() {
   }
 
   @Override
-  org.stellar.sdk.xdr.Operation.OperationBody toOperationBody() {
+  org.stellar.sdk.xdr.Operation.OperationBody toOperationBody(AccountConverter accountConverter) {
```
Shaptic, line 50: Forcing this parameter for operations that don't support muxed accounts is a little awkward to me as an API consumer, but I'm not familiar enough with Java patterns to suggest alternatives.

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -256,9 +260,9 @@ public TransactionEnvelope toEnvelopeXdr() {
      * who will use a sequence number. When build() is called, the account object's sequence number
      * will be incremented.
      */
-    public Builder(TransactionBuilderAccount sourceAccount, Network network) {
-      checkNotNull(sourceAccount, "sourceAccount cannot be null");
-      mSourceAccount = sourceAccount;
+    public Builder(AccountConverter accountConverter, TransactionBuilderAccount sourceAccount, Network network) {
+      mAccountConverter = checkNotNull(accountConverter, "accountConverter cannot be null");
```
Shaptic, line 264: Maybe a default to non-muxed (since we want opt-in) would be more convenient here?

src/test/java/org/stellar/sdk/OperationTest.java
```diff
@@ -69,7 +69,29 @@ public void testPaymentOperation() throws FormatException, IOException, AssetCod
...
         assertEquals(
                 "AAAAAQAAAAC7JAuE3XvquOnbsgv2SRztjuk4RoBVefQ0rlrFMMQvfAAAAAEAAAAA7eBSYbzcL5UKo7oXO24y1ckX+XuCtkDsyNHOp1n1bxAAAAAAAAAAAlQL5AA=",
-                operation.toXdrBase64());
+                operation.toXdrBase64(AccountConverter.enableMuxed()));
+    }
+
+    @Test
+    public void testMuxedPaymentOperation() throws FormatException, IOException, AssetCodeLengthInvalidException {
+        String source = "MA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJVAAAAAAAAAAAAAJLK";
+        String destination = "MDQNY3PBOJOKYZSRMK2S7LHHGWZIUISD4QORETLMXEWXBI7KFZZMKAAAAAAMV7V2XYGQO";
```
Shaptic, line 78: We should have a test case mixing muxed and non-muxed addresses in the same operation.

src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
```diff
@@ -84,8 +84,8 @@ public void testNewChallengeRejectsMuxedClientAccount() throws InvalidSep10Chall
           timeBounds
       );
       fail();
-    } catch (FormatException e) {
-      assertEquals("Version byte is invalid", e.getMessage());
+    } catch (InvalidSep10ChallengeException e) {
```
Shaptic, line 87: Isn't this an odd exception for this scenario?

src/main/java/org/stellar/sdk/AccountConverter.java
```diff
@@ -0,0 +1,41 @@
...
+    muxed.setEd25519(med25519.getEd25519());
+    return muxed;
+  }
+
+  public String decode(MuxedAccount account) {
+    if (this.enableMuxed || account.getDiscriminant().equals(CryptoKeyType.KEY_TYPE_ED25519)) {
+      return StrKey.encodeStellarMuxedAccount(account);
+    }
+
+    return StrKey.encodeStellarAccountId(StrKey.muxedAccountToAccountId(account));
```
Shaptic, line 39: Just noting here that this is the same behavior as in the [JS SDK](https://github.com/stellar/js-stellar-base/blob/1145458b490a1e3aa1275cd257b35010ea487381/src/util/decode_encode_muxed_account.js#L60-L71): force a fully-muxed account to a G-address when `enableMuxed = false` :heavy_check_mark:

Shaptic (approved):
I'm not a huge fan of the forced `AccountConverter` pattern, but it does make for a fairly clean way to switch between muxing support being on/off. Otherwise, just minor comments below.

src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
tamirms: a sep10 challenge can only be generated using non-muxed accounts. Hence `InvalidSep10ChallengeException` is thrown when trying to validate a challenge token which has a muxed account

src/main/java/org/stellar/sdk/AllowTrustOperation.java
tamirms: this method is only used internally. any method which does not have a `public` modifier is only visible within the SDK.

I just pushed another commit which ensures that, for all `public` methods that handle account strings, we have two implementations: one which allows you to provide your own account converter and another which uses `AccountConverter.disableMuxed()` by default

Issue #299, opened 2020-10-13
Add helpers to obtain Claimable Balance IDs


2opremio:
CAP 23 (as part of Stellar Protocol 14 and released with Horizon 1.9.0) , allows creating and claiming Claimable Balances.

Those balances are queried and claimed using their ID which can be obtained:

1. Before submitting the transaction creating the claim. The Claim ID is internally just a Hash which can be inferred even before Core accepts the transaction. A helper method would be useful in order to obtain this.
2. After the transaction creating the claim is submitted (i.e. in the transaction result). However, the claimable ID is pretty buried withing the result XDR, making it hard to obtain through the raw API result. A helper would come in handy.

(1) is already addressed in  https://github.com/stellar/go/pull/3122 , (2) still needs to be implemented

PR #349, opened 2021-06-24
Add getClaimableBalanceId() to determine claimable balance ids

tamirms:
Fixes https://github.com/stellar/java-stellar-sdk/issues/299

### DIFF
--- a/src/main/java/org/stellar/sdk/Transaction.java
+++ b/src/main/java/org/stellar/sdk/Transaction.java
@@ -95,15 +95,20 @@ public class Transaction extends AbstractTransaction {
   }
 
   /**
-   * Returns returns the claimable balance ID for the CreateClaimableBalanceOperation at the given index within the transaction.
+   * Returns the claimable balance ID for the CreateClaimableBalanceOperation at the given index within the transaction.
    */
   public String getClaimableBalanceId(int index) throws IOException {
-    if (index >= mOperations.length) {
+    if (index < 0 || index >= mOperations.length) {
       throw new IllegalArgumentException("index: " + index + " is outside the bounds of the operations within this transaction");
     }
     if (!(mOperations[index] instanceof CreateClaimableBalanceOperation)) {
       throw new IllegalArgumentException("operation at index " + index + " is not of type CreateClaimableBalanceOperation: " + mOperations[index].getClass());
     }
+
+    // We mimic the relevant code from Stellar Core
+    // https://github.com/stellar/stellar-core/blob/9f3cc04e6ec02c38974c42545a86cdc79809252b/src/test/TestAccount.cpp#L285
+    //
+    // Note that the source account must be *unmuxed* for this to work.
 
     OperationID id = new OperationID();
     id.setDiscriminant(EnvelopeType.ENVELOPE_TYPE_OP_ID);
@@ -122,7 +127,7 @@ public class Transaction extends AbstractTransaction {
 
     outputStream.reset();
     result.encode(new XdrDataOutputStream(outputStream));
-    return Util.bytesToHex(outputStream.toByteArray());
+    return Util.bytesToHex(outputStream.toByteArray()).toLowerCase();
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -92,6 +94,37 @@ public long getFee() {
     return mOperations;
   }
 
+  /**
+   * Returns returns the claimable balance ID for the CreateClaimableBalanceOperation at the given index within the transaction.
```
Shaptic, line 98: nit: "Returns returns"

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -92,6 +94,37 @@ public long getFee() {
...
+   */
+  public String getClaimableBalanceId(int index) throws IOException {
+    if (index >= mOperations.length) {
```
Shaptic, line 101: nit: do we need to check < 0 too? or `uint`ify the index

src/test/java/org/stellar/sdk/ClaimableBalanceIdTest.java
```diff
@@ -0,0 +1,105 @@
...
+        )).build();
+    Transaction transaction = new Transaction.Builder(AccountConverter.enableMuxed(), new Account(sourceAccount, 123l), Network.TESTNET)
+        .addOperation(op0)
+        .addOperation(new BumpSequenceOperation.Builder(2l).build())
+        .addOperation(op0)
+        .setTimeout(Transaction.Builder.TIMEOUT_INFINITE)
+        .setBaseFee(Transaction.MIN_BASE_FEE)
+        .build();
+
+    String expectedIdIndex0 = "0000000095001252AB3B4D16ADBFA5364CE526DFCDA03CB2258B827EDBB2E0450087BE51";
```
Shaptic, line 37: Why are they uppercase? I don't think `GET /claimable_balance/:id` is case-agnostic, so we'd want to make sure they're compatible.

src/main/java/org/stellar/sdk/Transaction.java
Shaptic: Can we also add a cross-reference to the docstring here? Either to the relevant [Stellar Core code](https://github.com/stellar/stellar-core/blob/9f3cc04e6ec02c38974c42545a86cdc79809252b/src/test/TestAccount.cpp#L285) or to the [txnbuild implementation](https://github.com/stellar/go/blob/master/txnbuild/transaction.go#L356). It'll be super helpful if we need to return to this code.

src/test/java/org/stellar/sdk/ClaimableBalanceIdTest.java
Shaptic: Either in the encoding or in the request builder, not sure what's more appropriate

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -92,6 +94,37 @@ public long getFee() {
...
+    body.setSeqNum(new SequenceNumber(new Int64(getSequenceNumber())));
+    body.setSourceAccount(AccountConverter.disableMuxed().encode(getSourceAccount()));
+    id.setId(body);
+
+    ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+    id.encode(new XdrDataOutputStream(outputStream));
+
+    ClaimableBalanceID result = new ClaimableBalanceID();
+    result.setDiscriminant(ClaimableBalanceIDType.CLAIMABLE_BALANCE_ID_TYPE_V0);
+    result.setV0(new Hash(Util.hash(outputStream.toByteArray())));
```
Shaptic, line 121: This is probably obvious with context & usage but there's no risk of the hash somehow... not being sha2 at some point?

Shaptic (approved):
A few minor notes, but LGTM otherwise! :+1:

src/main/java/org/stellar/sdk/Transaction.java
tamirms: unless we decide to to swap out sha2 for a different algorithm for transaction hashing it should remain as sha2

PR #359, opened 2021-09-17
AMM SDK Updates

paulbellamy:
Fixes https://github.com/stellar/java-stellar-sdk/issues/357

This adds the new functionality and updates for AMMs, into the java sdk.

I'm definitely not a Java expert, so feedback here would be great.

Stuff I'm not sure about:
- [ ] `Optional.absent` vs `null` for missing fields. In other fields we've used `null`, but `Optional.absent` seems cleaner? Not sure what the convention is here.
- [ ] Should we uppercase all `Id` instances to `ID`?
- [ ] Should we have a `hashCode` on everything?
- [ ] Should we have a `toString` on everything?
- [ ] Should `ChangeTrustAsset` and `TrustLineAsset` be sortable?


src/main/java/org/stellar/sdk/LiquidityPoolConstantProductParameters.java
```diff
@@ -0,0 +1,96 @@
...
+    return new LiquidityPoolConstantProductParameters(
+      Asset.fromXdr(xdr.getAssetA()),
+      Asset.fromXdr(xdr.getAssetB()),
+      xdr.getFee().getInt32().intValue()
+    );
+  }
+
+
+  @Override
+  public final org.stellar.sdk.LiquidityPoolID getId() {
```
tamirms, line 88: I think this class is in the same package as org.stellar.sdk.LiquidityPoolID so you should be able to omit the package prefix and have the method signature be:

`public final LiquidityPoolID getId()`

src/main/java/org/stellar/sdk/LiquidityPoolConstantProductParameters.java
```diff
@@ -0,0 +1,96 @@
...
+    return Objects.equal(this.getAssetA(), o.getAssetA())
+      && Objects.equal(this.getAssetB(), o.getAssetB())
+      && Objects.equal(this.getFee(), o.getFee());
+  }
+
+  /**
+   * Generates XDR object from a given LiquidityPoolParameters object
+   */
+  @Override
+  public final org.stellar.sdk.xdr.LiquidityPoolParameters toXdr() {
```
tamirms, line 60: similar to the comment below you can omit the "org.stellar.sdk.xdr" prefix everywhere in this file

src/main/java/org/stellar/sdk/Asset.java
```diff
@@ -32,6 +32,13 @@ public static Asset create(String type, String code, String issuer) {
     }
   }
 
+  public static Asset create(ChangeTrustAsset.Wrapper wrapped) {
```
tamirms, line 35: (wrapped)`

paulbellamy: Yeah, my thinking was that the conversions would be more symmetric this way. e.g. `ChangeTrustAsset.create(native)`, or `Asset.create(changeTrustAsset)`

src/main/java/org/stellar/sdk/LiquidityPoolConstantProductParameters.java
paulbellamy: I don't think we can here, because there's a collision with `LiquidityPoolParameters` in the current package.

tamirms: yeah, you're right. sorry you can ignore this comment

src/main/java/org/stellar/sdk/LiquidityPoolDepositOperation.java
```diff
@@ -0,0 +1,212 @@
...
+        private Optional<Asset> a = Optional.absent();
+        private Optional<Asset> b = Optional.absent();
+        private String maxAmountA;
+        private String maxAmountB;
+        private Price minPrice;
+        private Price maxPrice;
+
+        private String mSourceAccount;
+
+        Builder(LiquidityPoolDepositOp op) {
```
tamirms, line 82: it seems like all these builder constructors should actually be constructors for LiquidityPoolDepositOperation. Each of these constructors has enough information from the parameters to construct a LiquidityPoolDepositOperation instance. Maybe we can move the constructors to `LiquidityPoolDepositOperation` and get rid of the builder entirely?

paulbellamy: I guess, it is better to be simple or consistent? Don't know what the java norm is.

tamirms: Usually builder constructors either don't have any parameters or have a subset of the full parameter set and allow you to provide the missing parameters via setters. In this case, I don't think we get any benefit from the builder because we require all the parameters up front

src/main/java/org/stellar/sdk/responses/LiquidityPoolResponse.java
```diff
@@ -0,0 +1,149 @@
...
+  @SerializedName("id")
+  private LiquidityPoolID id;
+  @SerializedName("paging_token")
+  private String pagingToken;
+  @SerializedName("fee_bp")
+  private Integer feeBP;
+  @SerializedName("type")
+  private LiquidityPoolType type;
+  @SerializedName("total_trustlines")
+  private String totalTrustlines;
```
tamirms, line 30: you can use Longs instead of Strings because the java json parser will deserialize the string properly into a long

src/main/java/org/stellar/sdk/responses/LiquidityPoolResponse.java
```diff
@@ -0,0 +1,149 @@
...
+  public Reserve[] getReserves() {
+    return reserves;
+  }
+
+  /**
+   * Represents liquidity pool reserves.
+   */
+  public static class Reserve {
+    @SerializedName("amount")
+    private final String amount;
```
tamirms, line 75: same comment about Long instead of String

src/main/java/org/stellar/sdk/responses/effects/LiquidityPool.java
```diff
@@ -0,0 +1,55 @@
...
+
+public class LiquidityPool {
+  @SerializedName("id")
+  protected final LiquidityPoolID id;
+  @SerializedName("fee_bp")
+  protected final Integer feeBP;
+  @SerializedName("type")
+  protected final LiquidityPoolType type;
+  @SerializedName("total_trustlines")
+  protected final Integer totalTrustlines;
```
tamirms, line 17: this should be a Long

src/main/java/org/stellar/sdk/responses/LiquidityPoolResponse.java
paulbellamy: But this is a float...

tamirms: yeah, sorry about that. I forgot that they're actual floating point values

src/main/java/org/stellar/sdk/responses/TradePrice.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.math.BigDecimal;
+import java.math.MathContext;
+
+/**
+ * Represents Price. Price in Stellar is represented as a fraction.
+ */
+public class TradePrice {
+    @SerializedName("n")
+    private final String n;
```
tamirms, line 14: @paulbellamy I think `n` and `d` could be Longs

tamirms (approved):

src/main/java/org/stellar/sdk/responses/effects/LiquidityPoolTradeEffectResponse.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Represents liquidity_pool_trade effect response.
+ *
+ * @see <a href="https://www.stellar.org/developers/horizon/reference/resources/effect.html" target="_blank">Effect documentation</a>
+ * @see org.stellar.sdk.requests.EffectsRequestBuilder
+ * @see org.stellar.sdk.Server#effects()
+ */
+public class LiquidityPoolTradeEffectResponse extends EffectResponse {
+  @SerializedName("liquidity_pool")
```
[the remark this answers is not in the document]
tamirms: @Kirbyrawr the field should be liquidity_pool but the type of the field is wrong. the fix is here https://github.com/stellar/java-stellar-sdk/pull/363

PR #363, opened 2021-09-23
Represent pool shares amount as String

tamirms:


bartekn (approved):
LGTM but let's wait for @paulbellamy approval.

Issue #369, opened 2021-10-12
Test method name incorrect in AssetsRequestBuilderTest


gcatanese:
version = '0.28.0'

In AssetsRequestBuilderTest the method `testAccounts` should be called `testAssets`

PR #370, opened 2021-10-12
#369 Rename test method

gcatanese:
In AssetsRequestBuilderTest the method was erroneously called `testAccounts`, it is now renamed


gcatanese:
Closing as the topic branch was not created

Issue #384, opened 2021-12-01
get effects exception


przemyslawcelinski:
### What version are you using?

0.29.0

### What did you do?

```
        Page<EffectResponse> effectsPage = server.effects()
            .forTransaction("c0ec22df912e5911ecd260046806b826bf3a8596ea5aefac29a92131f565ea6c")
            .execute();
```

### What did you expect to see?
Effects for transaction `c0ec22df912e5911ecd260046806b826bf3a8596ea5aefac29a92131f565ea6c`

### What did you see instead?
```
shadow.com.google.gson.JsonSyntaxException: java.lang.NumberFormatException: For input string: "18446744073709551615"
at shadow.com.google.gson.internal.bind.TypeAdapters$11.read(TypeAdapters.java:306)
at shadow.com.google.gson.internal.bind.TypeAdapters$11.read(TypeAdapters.java:296)
at shadow.com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.read(ReflectiveTypeAdapterFactory.java:131)
at shadow.com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:222)
at shadow.com.google.gson.Gson.fromJson(Gson.java:927)
at shadow.com.google.gson.Gson.fromJson(Gson.java:994)
at shadow.com.google.gson.Gson.fromJson(Gson.java:967)
at org.stellar.sdk.responses.EffectDeserializer.deserialize(EffectDeserializer.java:38)
at org.stellar.sdk.responses.EffectDeserializer.deserialize(EffectDeserializer.java:18)
at shadow.com.google.gson.internal.bind.TreeTypeAdapter.read(TreeTypeAdapter.java:69)
at shadow.com.google.gson.internal.bind.TypeAdapterRuntimeTypeWrapper.read(TypeAdapterRuntimeTypeWrapper.java:41)
at shadow.com.google.gson.internal.bind.CollectionTypeAdapterFactory$Adapter.read(CollectionTypeAdapterFactory.java:82)
at shadow.com.google.gson.internal.bind.CollectionTypeAdapterFactory$Adapter.read(CollectionTypeAdapterFactory.java:61)
at shadow.com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.read(ReflectiveTypeAdapterFactory.java:131)
[...]

PR #388, opened 2021-12-15
Use BigInteger for Muxed ID data type

sreuland:
Problem: the muxed id values from horizon API is a uint64, java was mapping these into `long` which is signed int64 data types which left possibility of overflow. 

Proposed Solution: to safely handle these muxed id values, use the BigInteger data type.

Closes: #384

### DIFF
--- a/src/test/java/org/stellar/sdk/responses/OperationDeserializerTest.java
+++ b/src/test/java/org/stellar/sdk/responses/OperationDeserializerTest.java
@@ -3,13 +3,14 @@ package org.stellar.sdk.responses;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import junit.framework.TestCase;
-
-import org.stellar.sdk.*;
-
 import org.junit.Test;
+import org.stellar.sdk.*;
 import org.stellar.sdk.responses.operations.*;
 
+import java.math.BigInteger;
 import java.util.Arrays;
+
+import static java.math.BigInteger.valueOf;
 
 public class OperationDeserializerTest extends TestCase {
   @Test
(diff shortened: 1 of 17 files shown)


src/test/java/org/stellar/sdk/responses/OperationDeserializerTest.java
```diff
@@ -46,7 +47,7 @@ public void testDeserializeCreateAccountOperation() {
 
     assertEquals(operation.getSourceAccount(), "GD6WU64OEP5C4LRBH6NK3MHYIA2ADN6K6II6EXPNVUR3ERBXT4AN4ACD");
     assertEquals(operation.getPagingToken(), "3936840037961729");
-    assertEquals(operation.getId(), new Long(3936840037961729L));
+    assertEquals(operation.getId().longValue(), 3936840037961729L);
```
sreuland, line 50: not related, removed compiler deprecation warnings.

Shaptic (approved):
LGTM, and I don't really know much about Java, but are we *sure* this is necessary? Some cursory, naive research [suggests not](https://stackoverflow.com/a/397997), but then again there is a direct exception that led to this...

(Again, I don't know much about Java, so this is probably just a learning opportunity for myself.)

sreuland:
> LGTM, and I don't really know much about Java, but are we _sure_ this is necessary? Some cursory, naive research [suggests not](https://stackoverflow.com/a/397997), but then again there is a direct exception that led to this...
> 
> (Again, I don't know much about Java, so this is probably just a learning opportunity for myself.)

interesting, I think BigInteger was mentioned later in the roll there, yeah, its usage is a trade-off for less code/complexity. In this sdk use case, these id's aren't used in math ops either, they are only used as opaque identities, so, string representation would seem fine as simple alternative too. The approach done with BigInteger was to stay within jvm standard lib and consistent with the numeric 'id' convention already followed in the sdk, maybe at later rev point, these 'id' can be refactored to string.

cc @tamirms

tamirms (approved):

PR #390, opened 2021-12-17
Make constructures public to allow creating objects without gson.

lijamie98:


src/main/java/org/stellar/sdk/responses/AccountResponse.java
```diff
@@ -393,7 +393,7 @@ public String get(String key) {
     @SerializedName("transactions")
     private final Link transactions;
 
-    Links(Link effects, Link offers, Link operations, Link self, Link transactions) {
+    public Links(Link effects, Link offers, Link operations, Link self, Link transactions) {
```
sreuland, line 396: @lijamie98 , was just curious what is the use case driving the need to populate the model manually rather than from server api responses?

sreuland:
@lijamie98 , would you mind adding a line here in [CHANGELOG](https://github.com/stellar/java-stellar-sdk/blob/master/CHANGELOG.md?plain=1#L10) to note modification? since it's client facing, thanks!

src/main/java/org/stellar/sdk/responses/AccountResponse.java
lijamie98: Here is where I use it
https://github.com/stellar/java-stellar-anchor-dev/pull/7/files#diff-bb19215491831165a984cf8e66a77832f8567401759bba44621d8a21ff46dbceR169

This is an work around. I had to create another class TestSigner to convert to Signer at:
https://github.com/stellar/java-stellar-anchor-dev/pull/7/files#diff-bb19215491831165a984cf8e66a77832f8567401759bba44621d8a21ff46dbceR42

lijamie98: It is actually the Signer class that I need a public constructor. However, similar pattern may occur when we do tests in the future. 

Builder pattern would work too if public constructors are not preferred. Changing to Builder pattern will also need to change the way we construct AccountResponse.

lijamie98:
Will do

src/main/java/org/stellar/sdk/responses/AccountResponse.java
sreuland: I see, it's for testing concerns. I noticed a few other response POJO's here in the SDK already had public constructors on their inner classes, so, hopefully this is contained to AccountResponse. I wanted to also mention, Spring's `ReflectionTestUtils` , as an optional brute force way for tests to access private members of classes under test. I didn't see Spring in the anchor project's pom.xml, but maybe could use that with test scope if this issue of access crops up more.

Yes, It might be nice to for us to consider adding Lombok here into the SDK on these POJOs, and bring in builders. I see you have that in the anchor project as well, nice!

lijamie98: One of the anchor dev project requirements is NOT to use Spring :-(. So, I used gson to bruteforce it. However, I think this can be better solved with either a builder or public constructor. 

Lombok is a real nice way to reduce the verboseness of Java. :-)  +1

I will change the CHANGELOG on Monday and update the PR.

lijamie98:
CHANGELOG.md is also updated.

tamirms (approved):

sreuland:
@lijamie98 , looks like the commit from your github acct doesn't have gpg signature enabled, the PR can't metge non signed commits in this repo, can you redo commit after gpg enabled? Or lmk, i can assist.

lijamie98:
Thanks for offering the assitance! The commit was signed but the email that I used to sign the PR was mismatching. 

Fixed the user.email config in my git. It is now work.

Issue #397, opened 2022-01-06
Support muxed addresses (SEP-23) by default.


Shaptic:
As part of the rollout of muxed accounts (addresses in the `M...` form which combine a `G...` address alongside a uint64 identifier), we need to make M-addresses supported by default. In #348 we enabled their support via explicit _opt-in_; now, **developers should _disable_ supporting them only via explicit _opt-out_**.

This is (intentionally) a breaking change.

As a reminder, the following fields can contain a muxed account. Note that while the below list should be accurate, the XDR is always an authoritative source:

 - the `SourceAccount` of _any_ `Operation`
 - the `Destination` of the three payment operations: `PaymentOp`, `PathPaymentStrictReceive`, and `PathPaymentStrictSend`
 - the `Destination` of an `AccountMergeOp`
 - the `From` of a `ClawbackOp`
 - the `SourceAccount` of _any_ `Transaction`
 - the `FeeSource` of a `FeeBumpTransaction`

PR #398, opened 2022-01-06
#396: streamlined Asset creation to be consistent for all asset types

### LINKED ISSUE #396: Call getAsset from TrustlineCreatedEffectResponse class
What version are you using?
Version 0.30.0

What did you do?
Call method

```
  public Asset getAsset() {
    if (assetType.equals("native")) {
      return new AssetTypeNative();
    } else {
      return Asset.createNonNativeAsset(assetCode, assetIssuer);
    }
  }
```
from `TrustlineCreatedEffectResponse` class

Operation "change_trust" in block 38115883 with effect `trustline_created`
https://horizon.stellar.org/transactions/a0a65ac4d71e44fc2be2bb071645a16a577efe6258e610214ab5dc60309c9798/effects which has
"asset_type": "liquidity_pool_shares" but assetCode and assetIssuer are null. So in Asset.createNonNativeAsset(assetCode, assetIssuer); is thrown NullPointerException

```
  public static Asset createNonNativeAsset(String code, String issuer) {
    if (code.length() >= 1 && code.length() <= 4) {
      return new AssetTypeCreditAlphaNum4(code, issuer);
    } else if (code.length() >= 5 && code.length() <= 12) {
      return new AssetTypeCreditAlphaNum12(code, issuer);
    } else {
      throw new AssetCodeLengthInvalidException();
    }
  }
```

I added comment to closed issue: https://github.com/stellar/java-stellar-sdk/issues/374#issuecomment-1004622229

sreuland:
checked for Asset representation(native, alpha4, alpha12, pool share) consistency for any API operation/effect resource that may have possibility of holding an Asset resource which refers to liquidity pool shares via `liquidity_pool_id` and does not contain the usual non-native alpha4 or alpha12 attributes of `asset_code` and `asset_issuer`, these are the resources identified:

`
TrustlineCreatedEffectReponse
TrustlineUpdatedEffectResponse
TrustlineRemovedEffectResponse
ChangeTrustOperationResponse
`

I refactored the Asset class to provide a bit more asset-type agnostic factory methods for generating Assets which derive the type(i.e. class model), so callers do not need to have same know-how.

closes #396

### DIFF
--- a/src/main/java/org/stellar/sdk/Asset.java
+++ b/src/main/java/org/stellar/sdk/Asset.java
@@ -17,67 +17,68 @@ public abstract class Asset implements Comparable<Asset> {
    */
   public static Asset create(String canonicalForm) {
     if (canonicalForm.equals("native")) {
-      return new AssetTypeNative();
+      return create(canonicalForm, null, null);
     }
     String [] parts = canonicalForm.split(":");
     if (parts.length != 2) {
       throw new IllegalArgumentException("invalid asset "+ canonicalForm);
     }
-    return Asset.createNonNativeAsset(parts[0], parts[1]);
+    return create(null, parts[0], parts[1]);
   }
 
   /**
-   * Creates Asset for Alpha4/Alpha5/Native
+   * Creates Asset for Alpha4/Alpha12/Native
    *
-   * @param type the type of asset can be 'native', 'alpha4', 'alpha12'
-   * @param code the asset code that conforms to type or null
-   * @param issuer the asset issuer the conforms to type or null
-   * @return
+   * @param type the type of asset. 'native' will generate its respective asset sub-class,
+   *             if null or any other value will attempt to derive the asset sub-class from code and issuer.
+   * @param code the asset code or null
+   * @param issuer the asset issuer or null
+   * @return Asset
    */
   public static Asset create(String type, String code, String issuer) {
     return create(type, code, issuer, null);
   }
 
   /**
-   * Creates Asset for Alpha4/Alpha5/Native/LiquidityPool
+   * Creates Asset for Alpha4/Alpha12/Native/LiquidityPool
    *
-   * @param type the type of asset can be 'native', 'alpha4', 'alpha12' or 'liquidity_pool_shares'
-   * @param code the asset code that conforms to type or null
-   * @param issuer the asset issuer the conforms to type or null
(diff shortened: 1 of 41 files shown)


sreuland:
@tamirms , this could wait, can go in after #397 is done

src/main/java/org/stellar/sdk/Asset.java
```diff
@@ -17,67 +17,68 @@
...
-   * @param liquidityPoolID provided only if type is 'liquidity_pool_shares'
+   * @param type the type of asset. 'native' and 'liquidity_pool_shares' will generate their respective asset sub-classes
+   *             null or any other value will attempt to derive the asset sub-class from code and issuer.
+   * @param code the asset code or null
+   * @param issuer the asset issuer or null
+   * @param liquidityPoolID required only if type is 'liquidity_pool_shares'
    * @return Asset
    */
   public static Asset create(String type, String code, String issuer, String liquidityPoolID) {
+    if (type == null) {
```
tamirms, line 53: what's the rationale for changing the behavior of `type` to allow null values?

previously: type the type of asset can be 'native', 'alpha4', 'alpha12' or 'liquidity_pool_shares'

tamirms: https://github.com/stellar/go/blob/d550cca61c47b1b0bd3f9cc4fe1350d6e69b0f68/services/horizon/internal/assets/main.go#L17-L21

horizon uses credit_alphanum4 and credit_alphanum12 instead of alpha4 and alpha12

sreuland: the alpha4, alpha12 type values were/are ignored, rather the asset code is introspected to determine the Asset class type, only type values of 'native' and 'liquidity_pool_shares' have effect on derived Asset class, otherwise, clients don't need to pass a specific 'alpha4, alpha12' type, for backwards compat, it's fine, if it's one of those values or null will result in the same introspection of asset code to determine Asset class.

sreuland: trying to stay with existing static factory method pattern, but one other creational pattern to consider, but thought it might be too much breaking change, is to remove the `create` methods, and replace with builder pattern, such as:
`Asset.newBuilder().withCode().withIssuer().withLiquidityPoolId().create()`

if you don't include any `withXyz()` then by default the builder creates a native.

tamirms (approved):

PR #413, opened 2022-03-21
Update XDR for Protocol 19, add encoding support for CAP-40 SignedPayload Signer

### LINKED ISSUE #410: Update StrKey implementation to support CAP-40 signed payloads
Java-specific issue for the parent, stellar/go#4262.

### LINKED ISSUE #412: Update API definitions to conform to Protocol 19

sreuland:
Update the Java SDK to have the new XDR schema for [Protocol 19](https://github.com/stellar/go/tree/horizon-protocol-19)

this change focuses on supporting  [CAP-40](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0040.md#xdr-changes) , namely the new Signed Payload Signer as new StrKey.

Update the StrKey encoding/decoding to include the new Signed Payload Signer, add more tests where signed payload signer could appear such as SetOptionsOperation and AccountEntry.

Closes #412 
Closes #410 

can ignore all changed files in `src/main/java/org/stellar/sdk/xdr`, they are generated code from `xdrgen` tool against the latest xdr schema files.

This is first half of Protocol 19 support, a follow-on PR will be coming for remaining Protocol 19 support of [CAP-21](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0021.md) which introduces new tx pre-conditions on of which is `extraSigners` which this new Signed Payload Signer can be used.


src/main/java/org/stellar/sdk/Predicate.java
```diff
@@ -34,9 +37,9 @@ public static Predicate fromXdr(org.stellar.sdk.xdr.ClaimPredicate xdr) {
       case CLAIM_PREDICATE_NOT:
         return new Not(fromXdr(xdr.getNotPredicate()));
       case CLAIM_PREDICATE_BEFORE_RELATIVE_TIME:
-        return new RelBefore(xdr.getRelBefore().getInt64());
+        return new RelBefore(xdr.getRelBefore().getDuration().getInt64());
```
Shaptic, line 40: I think we should update these predicates to take their respective data types, i.e. `xdr.Duration` instead of the native `long`. Wdyt?

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -228,9 +248,12 @@ public static Transaction fromV0EnvelopeXdr(TransactionV0Envelope envelope, Netw
 
   public static Transaction fromV1EnvelopeXdr(AccountConverter accountConverter, TransactionV1Envelope envelope, Network network) {
     int mFee = envelope.getTx().getFee().getUint32();
+    TimeBounds mTimeBounds = null;
     Long mSequenceNumber = envelope.getTx().getSeqNum().getSequenceNumber().getInt64();
     Memo mMemo = Memo.fromXdr(envelope.getTx().getMemo());
-    TimeBounds mTimeBounds = TimeBounds.fromXdr(envelope.getTx().getTimeBounds());
+    if (envelope.getTx().getCond().getDiscriminant().equals(PreconditionType.PRECOND_TIME)) {
+      mTimeBounds = TimeBounds.fromXdr(envelope.getTx().getCond().getTimeBounds());
+    }
```
Shaptic, line 256: necessary but not sufficient: there can also be timebounds on the `PRECOND_V2` type. maybe a helper is in order for a general way to retrieve timebounds given a tx?

src/main/java/org/stellar/sdk/SignedPayloadSigner.java
```diff
@@ -0,0 +1,45 @@
...
+        return accountId;
+    }
+
+    /**
+     * get the binary format of a stellar account id, a Ed25519 public key
+     * @return stellar account id in binary format
+     */
+    public byte[] getDecodedAccountId() {
+        return StrKey.decodeStellarAccountId(accountId);
+    }
```
Shaptic, line 35: maybe `getSigner` and `getRawSigner` are more descriptive names? I always have to think twice about which one of encode/decode corresponds to string/bytes

src/main/java/org/stellar/sdk/StrKey.java
```diff
@@ -49,6 +65,25 @@ public static String encodeStellarAccountId(AccountID accountID) {
...
+        try {
+            ByteArrayOutputStream record = new ByteArrayOutputStream();
+            DataOutputStream dataStream = new DataOutputStream(record);
+            dataStream.write(signedPayloadSigner.getDecodedAccountId());
+            dataStream.writeInt(signedPayloadSigner.getPayload().length);
+            dataStream.write(signedPayloadSigner.getPayload());
+            int padding = signedPayloadSigner.getPayload().length % 4 > 0 ? 4 - signedPayloadSigner.getPayload().length % 4 : 0;
+            dataStream.write(new byte[padding]);
+            char[] encoded = encodeCheck(VersionByte.SIGNED_PAYLOAD, record.toByteArray());
+            return String.valueOf(encoded);
```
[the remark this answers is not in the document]
leighmcculloch: Maybe we should make it explicit in SEP-23 that the signed payload specification matches the XDR. Would that help @Shaptic ?

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -1,7 +1,23 @@
...
+import org.stellar.sdk.xdr.PreconditionType;
+import org.stellar.sdk.xdr.Preconditions;
+import org.stellar.sdk.xdr.SequenceNumber;
+import org.stellar.sdk.xdr.TransactionEnvelope;
+import org.stellar.sdk.xdr.TransactionSignaturePayload;
+import org.stellar.sdk.xdr.TransactionV0;
+import org.stellar.sdk.xdr.TransactionV0Envelope;
+import org.stellar.sdk.xdr.TransactionV1Envelope;
+import org.stellar.sdk.xdr.Uint32;
+import org.stellar.sdk.xdr.XdrDataOutputStream;
```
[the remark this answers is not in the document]
leighmcculloch: Importing specifics has been preferred in most Java workplaces I've worked in, so I think it is fine, but maybe we could make the change across the whole repo in one fell swoop as a sort of refactor independent of functional changes. I don't feel strongly.

src/main/java/org/stellar/sdk/SignedPayloadSigner.java
leighmcculloch: `getDecodedAccountId` doesn't just return a decoded an account ID, it returns the raw payload without a discriminant telling us if it is an ed25519 or not, so this new function isn't easily extensible to other key types if we ever add other key types. Could we make the accessor `default` instead of `public` since it seems like this is for internal use only which will also make it easier for us to change it in the future? Should make naming a little less crucial too since it isn't part of the exported API.

sreuland: @leighmcculloch , yes, can do refactor for this internal use, but am thinking to set this up now for scalability on future, I could change this to use AccountID which has the discriminator on key type, instead of String for account and that removes the extra getter methods down to pure pojo:

```
private org.stellar.sdk.xdr.AccountID accountId

public SignedPayloadSigner(AccountID accountId, byte[] payload) 

...

public AccountID getAccountID()
public byte[] getPayload()

```

src/main/java/org/stellar/sdk/StrKey.java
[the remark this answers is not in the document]
Shaptic: Yeah I think that was a big reason why I coded it the way I did initially: the description seemed like it was specifying a particular (and bespoke) binary packing, but in reality it was just a direct correspondence to `base32(xdr(signer) + xdr(payload))`. It would definitely help to tie the concepts together in the SEP, @leighmcculloch :+1:

src/main/java/org/stellar/sdk/Predicate.java
sreuland: I wove that in to reference the xdr temporal types, I noted minor backward compat/breaking aspect of this change in changelog.md. [5bf518bfd24](https://github.com/stellar/java-stellar-sdk/pull/413/commits/5bf518bfd24895520b5e4f391ad71470eda584af)

src/main/java/org/stellar/sdk/Transaction.java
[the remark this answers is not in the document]
sreuland: I can trigger a reformat on imports across the repo, yes, style convention usually discourage `*` package imports in favor of explicit paths.

src/main/java/org/stellar/sdk/Transaction.java
sreuland: yes, I'll use that suggestion, will be in the next follow-on PR that adds the CAP21 support on those new conditions, I included all the xdr schema changes up front in this pr, but only did the cap40 changes above first, this part ideally was just trying to refactor to maintain present compile/test until then.

src/main/java/org/stellar/sdk/StrKey.java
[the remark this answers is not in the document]
sreuland: ah, good to know, I updated to use the xdr bindings for encoding, thanks!

src/main/java/org/stellar/sdk/Transaction.java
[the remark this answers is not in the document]
sreuland: I will do this in a separate PR, I tried locally, and it changes about 250 files, which would make this PR formidable, thanks!

src/main/java/org/stellar/sdk/SignedPayloadSigner.java
sreuland: I refactored to this suggestion [c44b1de9fbe6](https://github.com/stellar/java-stellar-sdk/pull/413/commits/c44b1de9fbe6eb0b5bd98fe22b7771082b185384#diff-c1307eaeda0dc6eea5e80ffcbbf35abc90e6d3c41cdde9607ac6e67f94c4a904)

src/main/java/org/stellar/sdk/Predicate.java
sreuland: @Shaptic , i backed off a bit on this change and kept the backwards compliant interface for Predicate also - [163b582c5d](https://github.com/stellar/java-stellar-sdk/pull/413/commits/163b582c5d7c1ba9cb11ff69ca626c517af7c1be)

Shaptic: good call - I think I didn't realize this was a user-facing abstraction layer so maintaining bwd compatibility makes sense

src/main/java/org/stellar/sdk/Signer.java
```diff
@@ -4,12 +4,15 @@
...
 import org.stellar.sdk.xdr.Uint256;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
  * Signer is a helper class that creates {@link org.stellar.sdk.xdr.SignerKey} objects.
  */
 public class Signer {
+    public static final int SIGNED_PAYLOAD_MAX_PAYLOAD_LENGTH = 64;
```
Shaptic, line 14: Ah yeah, I think this belongs above in `SignedPayloadSigner`

src/main/java/org/stellar/sdk/StrKey.java
```diff
@@ -49,6 +65,28 @@ public static String encodeStellarAccountId(AccountID accountID) {
...
+            xdrPayloadSigner.setPayload(signedPayloadSigner.getPayload());
+            xdrPayloadSigner.setEd25519(signedPayloadSigner.getAccountId().getAccountID().getEd25519());
+
+            ByteArrayOutputStream record = new ByteArrayOutputStream();
+            xdrPayloadSigner.encode(new XdrDataOutputStream(record));
+
+            char[] encoded = encodeCheck(VersionByte.SIGNED_PAYLOAD, record.toByteArray());
+            return String.valueOf(encoded);
+        } catch (Exception ex) {
+            throw new FormatException(ex.getMessage());
```
Shaptic, line 86: Why rethrow when you can just let it bubble up?

src/main/java/org/stellar/sdk/Transaction.java
Shaptic: a way to do it here actually would be to say instead `.notEquals(PreconditionType.PRECOND_NONE)`, since both other forms have a timebound condition :thinking:

src/main/java/org/stellar/sdk/xdr/Asset.java
```diff
@@ -79,44 +79,39 @@ public Asset build() {
...
-      case ASSET_TYPE_CREDIT_ALPHANUM4:
-        AlphaNum4.encode(stream, encodedAsset.alphaNum4);
-        break;
-      case ASSET_TYPE_CREDIT_ALPHANUM12:
-        AlphaNum12.encode(stream, encodedAsset.alphaNum12);
-        break;
-      case ASSET_TYPE_POOL_SHARE:
-        throw new RuntimeException("Invalid asset type");
-    }
+  //Xdrgen::AST::Identifier
```
Shaptic, line 82: any reason why the formatting here got weird? :eyes:

src/main/java/org/stellar/sdk/xdr/ClaimableBalanceID.java
```diff
@@ -4,18 +4,16 @@
...
+import java.io.IOException;
+
 // === xdr source ============================================================
 
 //  union ClaimableBalanceID switch (ClaimableBalanceIDType type)
 //  {
 //  case CLAIMABLE_BALANCE_ID_TYPE_V0:
 //      Hash v0;
-//  case CLAIMABLE_BALANCE_ID_TYPE_FROM_POOL_REVOKE:
-//      Hash fromPoolRevoke;
```
Shaptic, line 18: unrelated? :thinking: I don't remember seeing this in the Go XDR update

src/main/java/org/stellar/sdk/StrKey.java
sreuland: to muffle the checked exceptions which otherwise would require adding on a `throws` declaration on the method for each one, which then forces the callers to specifically handle it, it's mostly for caller convenience.

src/main/java/org/stellar/sdk/Signer.java
sreuland: ok, I had located validation here in the [Signer.signedPayload(SignedPayloadSigner)](https://github.com/stellar/java-stellar-sdk/pull/413/files#diff-35deaf3ac00513f2a942ad2b4eb0415b4461402d49365bd38d312ec283e53eb3R84) factory method which constructs the xdr SignerKey for payload signer from the payload.  The intent on `SignedPayloadSigner` was to be data transfer object, no logic, to your point, it could take on that role also of validation, but thought was keeping that logic closer to when it's used in SignerKey, let me know if you want to discuss more, thanks!

src/main/java/org/stellar/sdk/SignedPayloadSigner.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * constructor
+     *
+     * @param accountId - the xdr AccountID
+     * @param payload - the raw payload for a signed payload
+     */
+    public SignedPayloadSigner(AccountID accountId, byte[] payload) {
+        this.accountId = checkNotNull(accountId);
+        this.payload = checkNotNull(payload);
```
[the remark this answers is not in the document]
sreuland: related [commment here](https://github.com/stellar/java-stellar-sdk/pull/413/files#r832695790)

src/main/java/org/stellar/sdk/SignedPayloadSigner.java
```diff
@@ -0,0 +1,58 @@
...
+import org.stellar.sdk.xdr.Uint256;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+
+/**
+ * Data model for the <a href="https://github.com/stellar/stellar-protocol/blob/master/core/cap-0040.md#xdr-changes">signed payload signer </a>
+ */
+public class SignedPayloadSigner {
+    private AccountID accountId;
```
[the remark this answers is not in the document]
sreuland: yes, updated that in [1717767a293a](https://github.com/stellar/java-stellar-sdk/pull/413/commits/1717767a293a462658c29f5f006d9511523ef3d3)

src/main/java/org/stellar/sdk/xdr/ClaimableBalanceID.java
sreuland: I think so, I diff'd *.x from p19 branch and here and there was some slight changes, I recopied all the .x from p19 branch into the java sdk, re-ran the xdrgen on those. cleaned this up.

src/main/java/org/stellar/sdk/xdr/Asset.java
sreuland: none that I can see, to be sure, I recopied latest .x from p19 branch and then regen'd code `xdrgen -o ./src/main/java/org/stellar/sdk/xdr -l java -n org.stellar.sdk.xdr ./xdr/Stellar-types.x ./xdr/Stellar-SCP.x ./xdr/Stellar-overlay.x ./xdr/Stellar-ledger-entries.x ./xdr/Stellar-ledger.x ./xdr/Stellar-transaction.x`, this was the result.

src/main/java/org/stellar/sdk/StrKey.java
Shaptic: makes sense, java-isms.. :+1:

src/main/java/org/stellar/sdk/StrKey.java
```diff
@@ -177,20 +229,20 @@ public static String encodeSha256Hash(byte[] data) {
...
-                return StrKey.base32Encoding.encode(unencoded).toCharArray();
+                return base32Encoding.encode(unencoded).toCharArray();
             }
 
             // Why not use base32Encoding.encode here?
             // We don't want secret seed to be stored as String in memory because of security reasons. It's impossible
             // to erase it from memory when we want it to be erased (ASAP).
             CharArrayWriter charArrayWriter = new CharArrayWriter(unencoded.length);
-            OutputStream charOutputStream = StrKey.base32Encoding.encodingStream(charArrayWriter);
+            OutputStream charOutputStream = base32Encoding.encodingStream(charArrayWriter);
```
Shaptic, line 245: Errm, note the comment above this line? I think there's a reason we use `StrKey`? Maybe @tamirms knows more on this (or `git blame` it)

Shaptic (approved):
1 small note of clarification around encoding, otherwise LGTM :shipit:

sreuland:
> You do have validation in two places (signer key and strkey), and imo you'd never want a SignedPayloadSigner that wasn't valid

@Shaptic , good point, since it spread to both, best to encapsulate once in the shared data object, updated [c61a5d531371](https://github.com/stellar/java-stellar-sdk/pull/413/commits/c61a5d531371cdc78bb98acc0cb4755e14bcc3ed)

src/main/java/org/stellar/sdk/StrKey.java
sreuland: this comment was referring to the usage of `.encode()` vs `.encodingStream()`, the change here was just source tidy-up, `StrKey.base32Encoding` was redundant, as within `StrKey` class scope, it references the same static member declared here at [base32Encoding](https://github.com/stellar/java-stellar-sdk/pull/413/files/49b477d5c24682dbe3dfa52fed0f9882a3c38d36#diff-d3b42d063d6edd65d7df86ba56cad3a838bffa1662d0de914d563c6eec6f9da0R56)

PR #414, opened 2022-03-24
Protocol 19 - CAP-40 Signed Payload Signer

sreuland:
have reviewed these changes for CAP-40 Signed Payload Signer on [PR #413](https://github.com/stellar/java-stellar-sdk/pull/413). renamed source repo branch to 'protocol-19', triggered [413](https://github.com/stellar/java-stellar-sdk/pull/413) to close, opened this PR on new branch name.

> Update the Java SDK to have the new XDR schema for [Protocol 19](https://github.com/stellar/go/tree/horizon-protocol-19)
>
> this change focuses on supporting  [CAP-40](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0040.md#xdr-changes) , namely the new Signed Payload Signer as new StrKey.
>
>Update the StrKey encoding/decoding to include the new Signed Payload Signer, add more tests where signed payload signer could appear such as SetOptionsOperation and AccountEntry.

Closes #412 
Closes #410 

>can ignore all changed files in `src/main/java/org/stellar/sdk/xdr`, they are generated code from `xdrgen` tool against the latest xdr schema files.

>This is first half of Protocol 19 support, a follow-on PR will be coming for remaining Protocol 19 support of [CAP-21](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0021.md) which introduces new tx pre-conditions on of which is `extraSigners` which this new Signed Payload Signer can be used.

PR #415, opened 2022-03-24
CAP 21, Transaction Preconditions and Account sequence fields

### LINKED ISSUE #409: Protocol 19 SDK Support
There are four elements to updating SDKs to support Protocol 19:

- [x] stellar/go#4260
- [x] stellar/go#4262
- [x] stellar/go#4281
- [ ] Updating API definitions in SDKs to handle the updated schemas from (1) and (2).

The last item is blocked until the Horizon API is finalized (i.e. after the PRs are merged).

### LINKED ISSUE #411: Extend transaction building to allow CAP-21 preconditions
Java-specific issue for the parent, stellar/go#4281.

sreuland:
These are updates to the Java SDK to support new Transaction Preconditions, and Account Entry fields(track sequence number updates) from [CAP-21](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0021.md).

* Refactored Transaction.Builder out of Transaction.java into its own class file, TransactionBuilder.java. 
* Incorporated using new PreconditionsV2 and Transaction.Preconditions. 
* maintain backwards compatibility on building transactions, but marked for deprecation existing usages of TimeBounds and Timeout in favor of using PreconditionsV2.TimeBounds instead.

for review, recommend looking at client usages of new preconditions first to see the net-net:
[TransactionBuilderTest.java, lines 89-413](https://github.com/stellar/java-stellar-sdk/pull/415/files#diff-bc45b2543f19733dce09373a49d7c819898c47adade2fc5a97b591286152d81bR89-R413)

then, followup with looking at the updated builder.
[TransactionBuilder.java](https://github.com/stellar/java-stellar-sdk/pull/415/files#diff-338b40b80a77385ce8f94bb8c79269a588b87a112a370760184837e2a5345357)

Closes #411
Closes #409

This effort is second half of Protocol 19 support in Java SDK specific to CAP-21, a prior [PR #414](https://github.com/stellar/java-stellar-sdk/pull/414) was done already for support of [CAP-40](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0040.md) for the SignedPayload Signer and merged into `protocol_19` branch. 
The plan is to merge this second PR into `protocol_19` branch and then [use follow on PR #416](https://github.com/stellar/java-stellar-sdk/pull/416) to squash merge `protocol_19` once into master when ready.


sreuland:
I'm in-process of adding the last bit of support in CAP-21 for the new AccountEntryExtensionV3.SeqLedger/SeqTime fields on the Account model and be pushing that up fairly soon, but wanted to make this PR visible now with all the new preconditions in place which is the bulk of changes.

build.gradle
```diff
@@ -1,90 +1,72 @@
 buildscript {
-  ext.okhttpclientVersion="3.11.0"
```
sreuland, line 2: not related to cap-21, but upgraded to latest gradle build, was able to user leaner config for shadowing jar.

src/main/java/org/stellar/sdk/responses/AccountResponse.java
```diff
@@ -25,6 +25,10 @@
   private Long sequenceNumber;
   @SerializedName("subentry_count")
   private Integer subentryCount;
+  @SerializedName("sequence_ledger")
+  private Long sequenceUpdatedAtLedger;
+  @SerializedName("sequence_time")
+  private Long sequenceUpdatedAtTime;
```
Shaptic, line 31: This should be a timestamp, no?

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -48,18 +46,17 @@
...
           Network network
   ) {
     super(accountConverter, network);
     this.mSourceAccount = checkNotNull(sourceAccount, "sourceAccount cannot be null");
-    this.mSequenceNumber = checkNotNull(sequenceNumber, "sequenceNumber cannot be null");
+    this.mSequenceNumber = sequenceNumber;
     this.mOperations = checkNotNull(operations, "operations cannot be null");
     checkArgument(operations.length > 0, "At least one operation required");
-
+    this.preconditions = preconditions;
```
Shaptic, line 57: `this.mPreconditions`, to conform?

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -38,7 +36,7 @@
   private final long mSequenceNumber;
   private final Operation[] mOperations;
   private final Memo mMemo;
-  private final TimeBounds mTimeBounds;
+  private final PreconditionsV2 preconditions;
```
Shaptic, line 39: We shouldn't use the XDR type here, but rather a user-friendly wrapper like we do in the Go SDK (same way we use `long` over `xdr.SequenceNumber` etc. here, see tangential convo on Discord)

Shaptic: Plus, a transaction can contain V1 (timebounds only), V2 (timebounds + others), or even no preconditions, so this wouldn't capture the other cases.

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -90,10 +87,21 @@ public Memo getMemo() {
...
+    return preconditions;
+  }
+
+  /**
+   * @return TimeBounds, or null (representing no time restrictions, i.e. infinite)
    */
   public TimeBounds getTimeBounds() {
-    return mTimeBounds;
+    TimeBounds timeBounds = TimeBounds.fromXdr(preconditions.getTimeBounds());
+    if (timeBounds == null || (timeBounds.getMinTime() < 1 && timeBounds.getMaxTime() < 1)) {
```
Shaptic, line 101: This is a breaking behavioral change, right? What's the purpose of introducing the "return null if infinite timebounds"?

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -173,7 +181,7 @@ private TransactionV0 toXdr() {
     transaction.setSourceAccountEd25519(StrKey.encodeToXDRAccountId(this.mSourceAccount).getAccountID().getEd25519());
     transaction.setOperations(operations);
     transaction.setMemo(mMemo.toXdr());
-    transaction.setTimeBounds(mTimeBounds == null ? null : mTimeBounds.toXdr());
+    transaction.setTimeBounds(preconditions == null || preconditions.getTimeBounds() == null ? null : preconditions.getTimeBounds());
```
Shaptic, line 184: Preconditions should never be null, as timebounds are required, so then we can unconditionally do `preconditions.getTimebounds()`, I think.

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -248,11 +257,15 @@ public static Transaction fromV0EnvelopeXdr(TransactionV0Envelope envelope, Netw
...
     Long mSequenceNumber = envelope.getTx().getSeqNum().getSequenceNumber().getInt64();
     Memo mMemo = Memo.fromXdr(envelope.getTx().getMemo());
+    PreconditionsV2 preconditionsV2 = new PreconditionsV2();
+
     if (envelope.getTx().getCond().getDiscriminant().equals(PreconditionType.PRECOND_TIME)) {
-      mTimeBounds = TimeBounds.fromXdr(envelope.getTx().getCond().getTimeBounds());
+      preconditionsV2 = new PreconditionsV2.Builder().timeBounds(envelope.getTx().getCond().getTimeBounds()).build();
+    }
+    if (envelope.getTx().getCond().getDiscriminant().equals(PreconditionType.PRECOND_V2) && envelope.getTx().getCond().getV2().getTimeBounds() != null) {
+      preconditionsV2 = envelope.getTx().getCond().getV2();
```
Shaptic, line 268: I can't tell if this is a naming conflict or incorrect functionality. If the envelope has PRECOND_TIME, we want to create PRECOND_TIME. If it has PRECOND_V2, we want to create PRECOND_V2 (regardless of whether or not those V2s actually have timebounds).

src/main/java/org/stellar/sdk/TransactionBuilder.java
```diff
@@ -0,0 +1,253 @@
...
+                // with a range of genesis to (current time + timeout interval)
+                preconditions.setTimeBounds(buildTimeBounds(0L, timeoutTimestamp));
+            } else {
+                // an existing timebounds in precondition, so timeout is conveyed on the tx as the current timebounds
+                // min time to a max time of (current time + timeout interval)
+                preconditions.getTimeBounds().setMaxTime(new TimePoint(new Uint64(timeoutTimestamp)));
+            }
+        }
+
+        if (timeout == 0) {
```
Shaptic, line 189: should we compare to `TIMEOUT_INFINITE` here?

src/main/java/org/stellar/sdk/TransactionBuilder.java
```diff
@@ -0,0 +1,253 @@
...
+     * This method is using the machine system time (UTC), make sure it is set correctly.
+     *
+     * @param timeout Timeout in seconds.
+     * @return updated Builder
+     * @see TimeBounds
+     * @deprecated this method will be removed in upcoming releases, use <code>addPreconditions()</code> instead
+     * for more control over preconditions.
+     */
+    public TransactionBuilder setTimeout(long timeout) {
+        if (preconditions.getTimeBounds() != null && preconditions.getTimeBounds().getMaxTime().getTimePoint().getUint64() > 0) {
```
Shaptic, line 168: lol that second conditional is so gross... there's definitely room for a `getTimebounds()` helper somewhere that extracts from preconditions and returns a non-XDR structure

src/main/java/org/stellar/sdk/SequenceNumberStrategy.java
```diff
@@ -0,0 +1,26 @@
...
+
+public interface SequenceNumberStrategy {
+
+    /**
+     * Derives the desired sequence number for a newly built transaction.
+     *
+     * @param account the source account of transaction.
+     * @return the sequence number to be applied on new transaction.
+     */
+    long getSequenceNumber(TransactionBuilderAccount account);
```
Shaptic, line 11: Interesting idea :thinking: I'm curious if @leighmcculloch thinks this is in line with how payment channels would want to and/or expect to use this

Is `getNextSequenceNumber` more accurate, to imply that this is generally an impure callback?

src/main/java/org/stellar/sdk/SequentialSequenceNumberStrategy.java
```diff
@@ -0,0 +1,13 @@
...
+
+public class SequentialSequenceNumberStrategy implements SequenceNumberStrategy{
+    @Override
+    public long getSequenceNumber(TransactionBuilderAccount account) {
+        return account.getIncrementedSequenceNumber();
+    }
+
+    @Override
+    public void updateSourceAccount(long newSequnceNumber, TransactionBuilderAccount account) {
+        account.incrementSequenceNumber();
```
Shaptic, line 11: Hmm, shouldn't this set the account sequence number to `newSequenceNumber`? Incrementing might not be the same.

Shaptic (commented):
Haven't finished reviewing yet but added some of my thoughts on the current impl. below

src/main/java/org/stellar/sdk/responses/AccountResponse.java
sreuland: I thought this was emitted[ from Horizon API json here](https://github.com/stellar/go/pull/4294/files#diff-491f1f1cafca2952be807956ff510ad82f7283582bad566e15b3e3ad00f451beR47) as integral Epoch right?

src/main/java/org/stellar/sdk/SequentialSequenceNumberStrategy.java
sreuland: it could, unlikely that account changes state between these two callback methods, this was refactored from what Transaction.Builder.build() did, which was `account.getIncrementedSequenceNumber()` , build tx, then `account.incrementSequenceNumber()`

src/main/java/org/stellar/sdk/Transaction.java
sreuland: oh, I removed that, it enabled  `SEP10ChallengeTest.testReadChallengeTransactionInvalidNoTimeBounds()`  to pass that infinite range on the updated tx builder in this version, and `SEP10Challenge.java` was checking for null timebounds on transaction, but that was only possible from test earlier when it used the Transaction constructor directly to pass timebounds=null, and now the builder doesn't allow that, I deleted the reference in `SEP10Challenge.java` and related test case also, as it's obsoleted by the tx builder enforcing presence of timebounds.

src/main/java/org/stellar/sdk/Transaction.java
sreuland: yeah, I [wondered about that ](https://github.com/stellar/go/pull/4303/files#r834845705), I'll refactor conditions into non-xdr data transfer object. 

The V1/V2/None conditions emitted in tx xdr will be fleshed out just a little more, I've got a [todo there now](https://github.com/stellar/java-stellar-sdk/pull/415/files#diff-54693c2dd591263011d0ad426cbf1f937b377134d13ce83bcc96977d38270b6cR214)

src/main/java/org/stellar/sdk/TransactionBuilder.java
sreuland: yep, that'll go away when refactor the preconditions to be non-xdr java pojo instead

src/main/java/org/stellar/sdk/Transaction.java
sreuland: this is unmarshalling tx xdr to java tx, so, once I refactor java tx.preconditions to be java pojo model instead of xdr then regardless of incoming tx xdr extension types, whichever attributes are available would converge/overlay into the one 'preconditions' pojo model here.

src/main/java/org/stellar/sdk/Transaction.java
sreuland: created wrapper [TransactionPreconditions](https://github.com/stellar/java-stellar-sdk/pull/415/files#diff-5ae3e337d7c0699e36fe2fec0ebe63510d548b2057bd400dc80186b8104042a3) to proxy the xdr.Preconditions.

src/main/java/org/stellar/sdk/Transaction.java
sreuland: I've got builder/conditions/tx aligned mostly to this rule with latest commits, but, I ran into a couple existing tests that demonstrated building xdr tx's with no timebounds based on the notion of 'timeout' being set from builder as an alternative to timebounds, such as [testParseV0Transaction on master branch](https://github.com/stellar/java-stellar-sdk/blob/master/src/test/java/org/stellar/sdk/TransactionTest.java#L336), so, I've retained the ability to construct a tx with no/null timebounds to support those test cases which asserted on the serialized/base64 xdr.

src/test/java/org/stellar/sdk/TransactionBuilderTest.java
```diff
@@ -0,0 +1,489 @@
...
+
+        transaction.sign(source);
+
+        Transaction transaction2 = (Transaction)Transaction.fromEnvelopeXdr(AccountConverter.enableMuxed(), transaction.toEnvelopeXdr(), Network.TESTNET);
+
+        assertEquals(transaction, transaction2);
+    }
+
+    @Test
+    public void testBuilderTimeBounds() throws FormatException, IOException {
```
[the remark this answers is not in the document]
Shaptic: one test I'd like to see:
 - start with tx xdr with v2 preconditions containing timebounds 
 - unpack into object
 - repack and make sure it creates v1 preconditions w/ only timebounds

Shaptic (approved):
Some notes below, but LGTM from my perspective. I think we should also have @tamirms review for as Java expert, though.

src/main/java/org/stellar/sdk/TransactionBuilder.java
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        long sequenceNumber = sequenceNumberStrategy.getSequenceNumber(mSourceAccount);
+
+        Operation[] operations = new Operation[mOperations.size()];
+        operations = mOperations.toArray(operations);
+        Transaction transaction = new Transaction(
+                mAccountConverter,
+                mSourceAccount.getAccountId(),
+                operations.length * mBaseFee,
```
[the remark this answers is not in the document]
sreuland: you're right, i see [that doc'd on fees](https://developers.stellar.org/docs/glossary/fees/), hmm, it looks like sdk has been doing this adjust on tx fee for a while, back to at least 2019 history. @tamirms , can you confirm intent, it looks like we need to change this to set tx fee to just `mBaseFee` correct?

src/main/java/org/stellar/sdk/SequenceNumberStrategy.java
tamirms: historically we have tried to keep the API of the different SDKs as consistent as possible. Unless we plan to adopt a similar interface in the Go SDK I'd advise from deviating from making any changes to the API which significantly deviate from the Go SDK

src/main/java/org/stellar/sdk/TransactionBuilder.java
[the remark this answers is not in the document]
tamirms: the code here is correct and the Go SDK does the same https://github.com/stellar/go/blob/master/txnbuild/transaction.go#L851-L855

the transaction xdr requires you to pass in the total fee, not the base fee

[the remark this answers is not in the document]
sreuland: thanks @tamirms , I filed [new-docs  ticket to update those docs ](https://github.com/stellar/new-docs/issues/659) to state the same.

src/main/java/org/stellar/sdk/TransactionBuilder.java
```diff
@@ -0,0 +1,242 @@
...
+        if (timeout < 0) {
+            throw new RuntimeException("timeout cannot be negative");
+        }
+
+        mTimeoutSet = true;
+        if (timeout > 0) {
+            long timeoutTimestamp = System.currentTimeMillis() / 1000L + timeout;
+            if (mTimeBounds == null) {
+                mTimeBounds = new TimeBounds(0, timeoutTimestamp);
+            } else {
```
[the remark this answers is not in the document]
sreuland: `TimeBounds(min,max)` constructor has check on that, I added a test here to trigger as well, [testBuilderFailsWhenTimeoutLessThanTimeBoundsMinimum](https://github.com/stellar/java-stellar-sdk/pull/415/files#diff-bc45b2543f19733dce09373a49d7c819898c47adade2fc5a97b591286152d81bR370) verify from builder.

src/main/java/org/stellar/sdk/TransactionBuilder.java
```diff
@@ -0,0 +1,242 @@
...
+public class TransactionBuilder {
+    private final TransactionBuilderAccount mSourceAccount;
+    private final AccountConverter mAccountConverter;
+    private Memo mMemo;
+    private List<Operation> mOperations;
+    private Integer mBaseFee;
+    private Network mNetwork;
+    private SequenceNumberStrategy sequenceNumberStrategy;
+    private TransactionPreconditions mPreconditions;
+    private TimeBounds mTimeBounds;
```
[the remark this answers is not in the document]
sreuland: just hadn't fully converted all plumbing of old api into new preconditions, but was able to refactor here and remove `mTimeBounds` to use just the new preconditions object for timebounds reference internally, thanks

src/main/java/org/stellar/sdk/Transaction.java
```diff
@@ -90,10 +84,17 @@ public Memo getMemo() {
   }
   
   /**
-   * @return TimeBounds, or null (representing no time restrictions)
```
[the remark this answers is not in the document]
sreuland: tidy'd up the docs here.

src/main/java/org/stellar/sdk/SequentialSequenceNumberStrategy.java
sreuland: followed your suggestion, it demonstrates using the callback better. [4d11a21edab2](https://github.com/stellar/java-stellar-sdk/pull/415/commits/4d11a21edab2760f31d55f45e5de114a55da728a)

src/main/java/org/stellar/sdk/SequenceNumberStrategy.java
sreuland: I refactored the tx builder sequence number resolution to be functional and pretty close to the same interface that txnbuild provides of either 'derive from sourceAccount.seqNum incremented by 1' or 'use an explicit value'.

https://github.com/stellar/java-stellar-sdk/pull/415/files#diff-338b40b80a77385ce8f94bb8c79269a588b87a112a370760184837e2a5345357R110

sreuland: I've [reverted back to original tx builder interface](https://github.com/stellar/java-stellar-sdk/pull/415/commits/4c8f5bf7d7152724b234d25c5b9d6e15a6c1e982) for using source account to get sequence number, just to move the p19 support through. we can do enhanced sequence number interfaces as follow-on if needed for tx's related to offchain payment flow use cases.

PR #416, opened 2022-03-25
Protocol 19

### LINKED ISSUE #409: Protocol 19 SDK Support

### LINKED ISSUE #410: Update StrKey implementation to support CAP-40 signed payloads

### LINKED ISSUE #411: Extend transaction building to allow CAP-21 preconditions

### LINKED ISSUE #412: Update API definitions to conform to Protocol 19

sreuland:
These are updates to the Java SDK to support Protocol 19.

This PR merges the feature branch that has already gone through prior reviews for specifics on CAP-21 and CAP-40 changes here:

https://github.com/stellar/java-stellar-sdk/pulls?q=is%3Apr+is%3Aclosed+label%3A%22Protocol+19%22

Closes #410 
Closes #411 
Closes #409 
Closes #412


CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
 
 As this project is pre 1.0, breaking changes may happen for minor version bumps. A breaking change will get clearly notified in this log.
 
+## 0.32.0 (Pending)
```
Shaptic, line 5: Can we add something about gradle? I don't know anything about Java build envs but maybe a note here about the changes would be helpful

src/main/java/org/stellar/sdk/FeeBumpTransaction.java
```diff
@@ -113,21 +120,18 @@ public static class Builder {
...
-            inner.getOperations(),
-            inner.getMemo(),
-            inner.getTimeBounds(),
-            inner.getNetwork()
-        );
+        this.mInner = new TransactionBuilder(inner.accountConverter, new Account(inner.getSourceAccount(), inner.getSequenceNumber() - 1), inner.getNetwork())
+                .setBaseFee((int)inner.getFee())
+                .addOperations(Arrays.asList(inner.getOperations()))
+                .addMemo(inner.getMemo())
+                .addPreconditions(new TransactionPreconditions.TransactionPreconditionsBuilder().timeBounds(inner.getTimeBounds()).build())
```
Shaptic, line 132: Can the `final Transaction inner` have other preconditions that need to be added here?

src/main/java/org/stellar/sdk/KeyPair.java
```diff
@@ -227,7 +234,9 @@ public byte[] sign(byte[] data) {
 
   /**
    * Sign the provided data with the keypair's private key and returns {@link DecoratedSignature}.
-   * @param data
+   *
+   * @param data the data to sign, typically the tx hash
```
Shaptic, line 238:
```suggestion
   * @param data the data to sign
```
It's the envelope that gets signed, I think?

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -157,11 +162,6 @@ public static ChallengeTransaction readChallengeTransaction(String challengeXdr,
       throw new InvalidSep10ChallengeException("The transaction sequence number should be zero.");
     }
 
-    // verify that transaction has time bounds set, and that current time is between the minimum and maximum bounds.
```
Shaptic, line 160: Why gone?

src/main/java/org/stellar/sdk/TransactionBuilder.java
```diff
@@ -0,0 +1,231 @@
...
+    public TransactionBuilder setTimeout(long timeout) {
+        if (mPreconditions.getTimeBounds() != null && mPreconditions.getTimeBounds().getMaxTime() != TransactionPreconditions.TIMEOUT_INFINITE) {
+            throw new RuntimeException("TimeBounds.max_time has been already set - setting timeout would overwrite it.");
+        }
+
+        if (timeout < 0) {
+            throw new RuntimeException("timeout cannot be negative");
+        }
+
+        mTimeoutSet = true;
```
Shaptic, line 161: Is this variable used elsewhere? Why isn't it checked on line 153 above?

src/main/java/org/stellar/sdk/TransactionBuilder.java
```diff
@@ -0,0 +1,231 @@
...
+        if (timeout > 0) {
```
Shaptic, line 162: sanity check: there's no `else` case, so will there be no preconditions built if timeout is set to inf?

src/main/java/org/stellar/sdk/TransactionPreconditions.java
```diff
@@ -0,0 +1,127 @@
...
+    public static TransactionPreconditions fromXdr(Preconditions preconditions) {
+        TransactionPreconditionsBuilder builder = new TransactionPreconditionsBuilder();
+
+        if (preconditions.getDiscriminant().equals(PreconditionType.PRECOND_V2)) {
+            if (preconditions.getV2().getTimeBounds() != null) {
+                builder.timeBounds(new TimeBounds(
+                        preconditions.getV2().getTimeBounds().getMinTime().getTimePoint().getUint64(),
+                        preconditions.getV2().getTimeBounds().getMaxTime().getTimePoint().getUint64()));
+            }
+            if (preconditions.getV2().getExtraSigners() != null && preconditions.getV2().getExtraSigners().length > 0) {
```
Shaptic, line 65: some/most of these can never be null, afaict

Shaptic (commented):
some minor notes :arrow_down:

CHANGELOG.md
sreuland: could mention, but the gradle version upgrade is not an external facing change, there is no developer touchpoint on it, it ends up here as pre-downloaded bins of the new version that developers still run with same cli `gradlew build|test`, they don't need to do additional host system changes for the gradle upgrade.

sreuland: could mention, but the gradle version upgrade is not an external facing change, there is no developer touchpoint on it, it ends up here as pre-downloaded bins of the new version that developers still run with same cli `gradlew build|test`, they don't need to do additional host system changes for the gradle upgrade.

src/main/java/org/stellar/sdk/FeeBumpTransaction.java
```diff
@@ -113,21 +120,18 @@ public static class Builder {
...
-            inner.accountConverter,
-            inner.getSourceAccount(),
-            inner.getFee(),
-            inner.getSequenceNumber(),
-            inner.getOperations(),
-            inner.getMemo(),
-            inner.getTimeBounds(),
-            inner.getNetwork()
-        );
+        this.mInner = new TransactionBuilder(inner.accountConverter, new Account(inner.getSourceAccount(), inner.getSequenceNumber() - 1), inner.getNetwork())
```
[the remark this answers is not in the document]
sreuland: shouldn't be, kept the interface of this FeeBumpTransaction.Builder the same, the parameters are one-way inbound, doesn't mutate state of those, the new local instance of account is created here to use the new TransactionBuilder interface, but the seqnum increment change that new builder applies on that account instance are ignored here, since that info was never propagated outward.

src/main/java/org/stellar/sdk/FeeBumpTransaction.java
sreuland: it's a `V0` tx since it's in this block, so TimeBounds is the only precondition'able attribute available from that version of tx right?

src/main/java/org/stellar/sdk/KeyPair.java
sreuland: well, the [hash of the serialized tx xdr and networkid](https://github.com/stellar/java-stellar-sdk/blob/protocol_19/src/main/java/org/stellar/sdk/Transaction.java#L66-L70),  but that was not changed here, I just added a little color there, and am fine removing this per suggestion.

src/main/java/org/stellar/sdk/Sep10Challenge.java
sreuland: looked redundant, the test will fail on L165 if it were null

src/main/java/org/stellar/sdk/TransactionBuilder.java
sreuland: thanks for drilling into that, reviewed this area and cleaned up considerably, was able to remove this usage.

src/main/java/org/stellar/sdk/TransactionBuilder.java
sreuland: yes, good catch on the edge case, I've reviewed this and cleaned up and added more test coverage on setting timeouts.

src/main/java/org/stellar/sdk/TransactionPreconditions.java
sreuland: there is no enforcement of null check in the generated xdr java stub classes, an instance of the stub class, xdr.PreconditionsV2 could be constructed with the builder which doesn't require any of these attribs to be set, which results in nulls:

https://github.com/stellar/java-stellar-sdk/pull/416/files#diff-1c2812de0a0b421d83904a598bacc4351a3c80d9701f4d7260ca9cf1b4f7ad00R197

Shaptic (approved):
:tada:

src/test/java/org/stellar/sdk/SetOptionsOperationTest.java
```diff
@@ -0,0 +1,43 @@
...
+import org.stellar.sdk.xdr.SignerKey;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+
+public class SetOptionsOperationTest {
+    KeyPair source = KeyPair.fromSecretSeed("SC4CGETADVYTCR5HEAVZRB3DZQY5Y4J7RFNJTRA6ESMHIPEZUSTE2QDK");
+
+    @Test
+    public void testPaylodSignerKey() {
```
Kirbyrawr, line 14: There is a typo here it should be `testPayloadSignerKey`

PR #417, opened 2022-04-08
payload signer signatures

sreuland:
As part of CAP-40, needed to include payload signer signature generation from `KeyPair` and separate way to add pre-built signatures onto a `Transaction`.

### DIFF
--- a/src/main/java/org/stellar/sdk/KeyPair.java
+++ b/src/main/java/org/stellar/sdk/KeyPair.java
@@ -9,8 +9,14 @@ import net.i2p.crypto.eddsa.spec.EdDSANamedCurveSpec;
 import net.i2p.crypto.eddsa.spec.EdDSANamedCurveTable;
 import net.i2p.crypto.eddsa.spec.EdDSAPrivateKeySpec;
 import net.i2p.crypto.eddsa.spec.EdDSAPublicKeySpec;
-
-import org.stellar.sdk.xdr.*;
+import org.stellar.sdk.xdr.DecoratedSignature;
+import org.stellar.sdk.xdr.PublicKey;
+import org.stellar.sdk.xdr.PublicKeyType;
+import org.stellar.sdk.xdr.SignatureHint;
+import org.stellar.sdk.xdr.SignerKey;
+import org.stellar.sdk.xdr.SignerKeyType;
+import org.stellar.sdk.xdr.Uint256;
+import org.stellar.sdk.xdr.XdrDataOutputStream;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
@@ -21,6 +27,7 @@ import java.security.SignatureException;
 import java.util.Arrays;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static java.lang.System.arraycopy;
 
 /**
  * Holds a Stellar keypair.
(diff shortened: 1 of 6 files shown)


src/main/java/org/stellar/sdk/KeyPair.java
```diff
@@ -241,6 +250,34 @@ public DecoratedSignature signDecorated(byte[] data) {
...
+    if (signerPayload.length >= hint.length) {
+      arraycopy(signerPayload, signerPayload.length - hint.length, hint, 0, hint.length);
+    } else {
+      arraycopy(signerPayload, 0, hint, 0, signerPayload.length);
+    }
+
+    //XOR the new hint with this keypair's public key hint
+    for (int i = 0; i < hint.length; i++) {
+      hint[i] ^= (i < payloadSignature.getHint().getSignatureHint().length ? payloadSignature.getHint().getSignatureHint()[i] : 0);
+    }
```
leighmcculloch, line 276: ❓ I'm a bit confused by this. It looks like there's a condition here to handle the `payloadSignature` having a too short hint, however the too short hint is handled a few lines above already. Hints are _always_ 4 bytes. A `SignatureHint` should never contain a hint shorter.

src/test/java/org/stellar/sdk/KeyPairTest.java
```diff
@@ -86,4 +90,35 @@ public void testSignWithoutSecret() {
...
+    // the hint from this keypair is [254,66,4,55]
+
+    byte[] payload = new byte[]{1,2,3};
+    DecoratedSignature sig = keypair.signPayloadDecorated(payload);
+    // the hint could only be derived off of 3 bytes from payload
+    Assert.assertArrayEquals(sig.getHint().getSignatureHint(), new byte[]{
+            (byte)(255),
+            (byte)(0xFF & 64),
+            (byte)(0xFF & 7),
+            (byte)(0xFF & 55)});
```
leighmcculloch, line 121: ❓ What is the 0xFF here for? Isn't 0xff & 64 = 0x64? I think you can simply do:
```suggestion
            (byte)(255),
            (byte)(64),
            (byte)(7),
            (byte)(55)});
```
See this as an example: https://go.dev/play/p/qi-HubnfmPA

leighmcculloch (commented):
Two questions (❓).

src/main/java/org/stellar/sdk/KeyPair.java
sreuland: `payloadSignature.getHint().getSignatureHint()` refers to the key hint, it was paranoid defensive guard on that key hint array being shorter than the payload hint array, fine with removing it for cleaner approach here.

leighmcculloch: Yeah I think you can remove the check. Hint's should always be 4 bytes. In other languages where we can guarantee the length of the byte[] is 4 we do that, I guess in Java we can't?

src/test/java/org/stellar/sdk/KeyPairTest.java
sreuland: yes, wasn't necessary, wanted to show the 8 bit unsigned value being placed into each byte, and did used that form for all values even the ones below 128

leighmcculloch (approved):

src/main/java/org/stellar/sdk/KeyPair.java
sreuland: java array type doesn't include the dimension as part of the type declaration and the generated xdr.SignatureHint doesn't have any checks on the size. it would require decorator or other wrapper class in the sdk around xdr.SignatureHint that enforces the length dimension. I [removed that guard check](https://github.com/stellar/java-stellar-sdk/pull/417/commits/e6e196609085d48198931ac1e6d87c4e50bd5797).

PR #425, opened 2022-05-01
Handle null/empty memo hash. Fix #392

### LINKED ISSUE #392: memo exception
0.29.0

### What did you do?

I try call getMemo for TransactionResponse
`transactionResponse.getMemo()`

result of memo from horizon https://horizon.stellar.org/transactions/b8084d1cf9b44ac9b84b4064a6b0919ed9251b1ead000a89e7abd483d1165b17
`"memo": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=",`

### What did you see instead?
`Method threw 'java.lang.ArrayIndexOutOfBoundsException' exception. Cannot evaluate org.stellar.sdk.MemoHash.toString()`

vinamogit:
It is possible to create a MemoHash with 32-byte array of zero byte.
Here this kind of hash is turned into the empty string, the hex value is 0x0000000000000000000000000000000000000000000000000000000000000000

Closes #392


vinamogit:
Fix #392

sreuland:
looks like worthwhile guard in any case, default to empty string rather than stack exception is a better outcome, @leighmcculloch , do you see any potential mis-representation of memo by defaulting null/empty xdr of memo to empty string java binding?

src/test/java/org/stellar/sdk/MemoTest.java
```diff
@@ -74,7 +74,41 @@ public void testParseMemoId() {
         MemoId memoId = (MemoId)transactionResponse.getMemo();
         assertEquals(longId, Long.toUnsignedString(memoId.getId()));
     }
+    
+    @Test
+    public void testParseMemoHash() {
```
sreuland, line 79: I don't think need this test case as it's going through code outside the target scope, `testMemoNullHexHashSuccess` exercises the same target code path correct?

vinamogit: I've added this one since it's the initial case from the bug report and what can be found on the blockchain.
testMemoNullHexHashSuccess shows that it is possible to create such a memo.

sreuland: Ok, in that case, it's probably worth re-locate this unit test to that target, namely the `TransactionDeserializerTest`, looks like there are a few test cases on memo already, could drop this one there also:

https://github.com/vinamogit/java-stellar-sdk/blob/fix-empty-memohash-exception/src/test/java/org/stellar/sdk/responses/TransactionDeserializerTest.java#L93

sreuland (approved):

src/main/java/org/stellar/sdk/MemoHashAbstract.java
```diff
@@ -54,7 +54,11 @@ public String getHexValue() {
    * </code>
    */
   public String getTrimmedHexValue() {
-    return this.getHexValue().split("00")[0];
+      String[] strArray = this.getHexValue().split("00");
+      if (strArray.length > 0) {
+          return strArray[0];
```
leighmcculloch, line 59: I think the bug here is that we attempt to trim the hash value at the first 0x00. Memo hashes are always 32bytes, fixed size, and so we should always return the full 32byte hash without truncating. By trimming at the first 0x00 we might be losing non-zero bytes that follow.

Shaptic: I think what you're saying, @leighmcculloch, is that `getTrimmedHexValue()` shouldn't exist, because the very notion of trimming a memo hash is incorrect (see docstring). There must be a historical reason for this helper, then :thinking:

Shaptic: Or, alternatively, this should do a `stripEnd("00")`? Not sure if that [usage](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#stripEnd-java.lang.String-java.lang.String-) is correct.

leighmcculloch: Yes we should remove getTrimmedHexValue. If an application is using a memo hash in this way, they should be using a different type of memo, or doing the truncation in their application. There is really no way that the Java SDK can know if a zero is safe to truncate, because zero is a significant data point in a hash.

leighmcculloch: +1 Applications can use the hash as a null terminated string if they choose, but that logic should live in the application, not in the SDK.

src/test/java/org/stellar/sdk/MemoTest.java
```diff
@@ -122,18 +144,21 @@ public void testMemoHashInvalidHex() {
         try {
             Memo.hash("test");
             fail();
-        } catch (Exception e) {
-
+        } catch (MemoWrongSizeException e1) {
+            fail();
+        } catch (IllegalArgumentException e2) {
```
sreuland, line 149: It's not immediately clear where this IAE is thrown from? I kinda expected MemoWrongSizeException in this case?

src/main/java/org/stellar/sdk/MemoWrongSizeException.java
```diff
@@ -0,0 +1,23 @@
+package org.stellar.sdk;
+
+/**
+ * Indicates that value passed to Memo has not the right size
+ * @see Memo
+ */
+public class MemoWrongSizeException extends RuntimeException {
```
leighmcculloch, line 7:
```suggestion
public class IncorrectMemoSizeException extends RuntimeException {
```

src/main/java/org/stellar/sdk/MemoWrongSizeException.java
```diff
@@ -0,0 +1,23 @@
...
+        this.size = length;
+    }
+
+    public MemoWrongSizeException(String message, int length) {
+        super(message);
+        this.size = length;
+    }
+    
+    public int getActualSize() {
+        return size;
```
leighmcculloch, line 21: We should name all the variables containing the actual size a consistent name. It reduces the concepts the reader needs to digest.

```suggestion
    private int actualSize;
    
    public MemoWrongSizeException(int actualSize) {
        super();
        this.actualSize = actualSize;
    }

    public MemoWrongSizeException(String message, int actualSize) {
        super(message);
        this.actualSize = actualSize;
    }
    
    public int getActualSize() {
        return actualSize;
```

src/test/java/org/stellar/sdk/MemoTest.java
```diff
@@ -112,8 +121,21 @@ public void testMemoHashTooLong() {
...
-            assertTrue(exception.getMessage().contains("MEMO_HASH can contain 32 bytes at max."));
+        } catch (MemoWrongSizeException exception) {
+            assertTrue(exception.getMessage().contains("MEMO_HASH must contain 32 bytes."));
+            assertEquals(33, exception.getActualSize());
+        }
+    }
+    
+    @Test
+    public void testMemoHashTooShort() {
+
```
leighmcculloch, line 132:
```suggestion
```

src/main/java/org/stellar/sdk/MemoText.java
```diff
@@ -23,7 +23,7 @@ public MemoText(XdrString text) {
     this.text = checkNotNull(text, "text cannot be null");
     int length = this.text.getBytes().length;
     if (length > 28) {
-      throw new MemoTooLongException("text must be <= 28 bytes. length=" + String.valueOf(length));
+      throw new MemoWrongSizeException("text must be <= 28 bytes. length=" + String.valueOf(length), length);
```
leighmcculloch, line 26: We should keep the too long exception for memo text. Otherwise, this breaks backwards compatibility for code that is outside the area affected by the bug we are fixing.

leighmcculloch (commented):
This is great. Thanks for updating the PR. There's a couple small changes I request inline. I also share @sreuland question about the IllegalArgumentException. Once they're addressed I think we can merge this.

src/test/java/org/stellar/sdk/MemoTest.java
vinamogit: I thought the same at first but since we are parsing the string to decode it to bytes, it makes sense to have the IAE.

sreuland: I think the specialized Memo exceptions are unnecessary granularity for clients to juggle, the excepts are un-checked, `RuntimeException`, which means callers can't recover from different types of this error and try something else, if that was the case then the Memo exceptions should extend `Exception` instead and be thrown from methods as checked, forcing callers to catch that, but since that is not applicable here, their granularity can just be expressed as different messages using the standard `RuntimeException`.

I would suggest to remove MemoWrongSizeException and MemoTooLongException , and just throw `new RuntimeException("memo problem explained here")` where needed, then this test just verifies `catch (Exception)` across all bad inputs, callers can't do something different depending on the type of invalid memo here, they have to move on and the msg provides context for logging purposes.

Thoughts otherwise?

There's a minor breaking change there on removing existing MemoTooLongException, I think it's worthwhile as it's less code for clients to manage, we have to mention it in the CHANGELOG which we do at release prep time, but if you want to add a note under new `pending` section at top of CHANGELOG.md, we can roll that in during the next release notes.

PR #426, opened 2022-05-01
Basic implementation of liquidity_pools?account

### LINKED ISSUE #376: Feature Request: Add request for fetching liquidity pools by account
In https://github.com/stellar/go/pull/4043 we added a new endpoint to horizon:

`GET /liquidity_pools?account=<id>` returns a list of pools that have as a participant

We should add support for the Java SDK to exercise this endpoint.

vinamogit:
Fixes #376


src/test/java/org/stellar/sdk/requests/LiquidityPoolsRequestBuilderTest.java
```diff
@@ -26,4 +26,13 @@ public void testForReserves() {
         .buildUri();
     assertEquals("https://horizon-testnet.stellar.org/liquidity_pools?reserves=EURT%3AGAP5LETOV6YIE62YAM56STDANPRDO7ZFDBGSNHJQIYGGKSMOZAHOOS2S%2CPHP%3AGAP5LETOV6YIE62YAM56STDANPRDO7ZFDBGSNHJQIYGGKSMOZAHOOS2S", uri.toString());
   }
+  
+  @Test
+  public void testForAccount() {
```
sreuland, line 31: I know it's not established here yet on other tests, but might be worthwhile to include coverage of an additional test for clearing the parameter from the builder, i.e.`LiquidityPoolsRequestBuilder.forAccount(null)` is valid, what would be the expected url if that is called or called in succession after first doing `.forAccount("xyz")`an allowed.

sreuland:
Looks good to me, thanks for contributing! @tamirms per your #376 , do you agree?

src/test/java/org/stellar/sdk/requests/LiquidityPoolsRequestBuilderTest.java
vinamogit: Will do that. I tried with forReserves(String...) but of course there is a NPE on forReserves(null).
Does a pass is required on this kind of method call?

sreuland: `forReserves(String ...)`interface prevents null parameter, might be worth a bit more coverage to assert a second call with different values for any of these builder methods, and verify if result is supposed to append or reset, It depends on `okhttp3.HttpUrl.Builder`, then can merge, as the PR looks good.

tamirms (approved):

overcat:
This PR seems to have been forgotten, I think we can merge it. @sreuland

PR #428, opened 2022-05-04
Include new Protocol 19 Preconditions on Transaction API resource

sreuland:
Update the Transaction API Response model to include new P19 Preconditions, this also supports the [latest changes in Horizon API ](https://github.com/stellar/go/pull/4361)to Preconditions json model of datetime attribs being expressed in unix epoch.

### DIFF
--- a/src/main/java/org/stellar/sdk/responses/TransactionResponse.java
+++ b/src/main/java/org/stellar/sdk/responses/TransactionResponse.java
@@ -2,7 +2,6 @@ package org.stellar.sdk.responses;
 
 import com.google.common.base.Optional;
 import com.google.gson.annotations.SerializedName;
-
 import lombok.Value;
 import org.stellar.sdk.Memo;
 
@@ -181,7 +180,7 @@ public class TransactionResponse extends Response implements Pageable {
     @SerializedName("ledgerbounds")
     LedgerBounds ledgerBounds;
     @SerializedName("min_account_sequence")
-    long minAccountSequence;
+    Long minAccountSequence;
     @SerializedName("min_account_sequence_age")
     long minAccountSequenceAge;
     @SerializedName("min_account_sequence_ledger_gap")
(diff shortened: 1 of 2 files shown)


src/main/java/org/stellar/sdk/responses/TransactionResponse.java
```diff
@@ -164,6 +171,41 @@ public Links getLinks() {
...
+   * Preconditions of a transaction per <a href="https://github.com/stellar/stellar-protocol/blob/master/core/cap-0021.md#specification">CAP-21<a/>
+   */
+  @Value
+  public static class Preconditions {
+    @SerializedName("timebounds")
+    TimeBounds timeBounds;
+    @SerializedName("ledgerbounds")
+    LedgerBounds ledgerBounds;
+    @SerializedName("min_account_sequence")
+    long minAccountSequence;
```
leighmcculloch, line 184: There is a distinction between when `minAccountSequence` is set or not, so you can't treat not set to a default value like 0. Not set means one-less than the tx seq, but 0 means 0. @Shaptic This may be an issue in Horizon's API and in other SDKs.

leighmcculloch: See related conversation: https://discord.com/channels/897514728459468821/971454936615358485/971454986988957736

sreuland: ah, ok, thanks for mentioning this, I updated the sdk here to model that as a nullable object to signal if wasn't present in the json.

Shaptic (approved):
Cool! Does the account response resource also have the new `sequence_ledger`/`time` fields?

src/main/java/org/stellar/sdk/responses/TransactionResponse.java
Shaptic: It's a pointer in the API, so we're good

sreuland:
> Cool! Does the account response resource also have the new `sequence_ledger`/`time` fields?

Yes, those have been added as nullable:
https://github.com/stellar/java-stellar-sdk/blob/master/src/main/java/org/stellar/sdk/responses/AccountResponse.java#L28-L31

PR #435, opened 2022-05-26
Add memo field to Sep10Challenge.newChallenge()

lijamie98:


src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -99,6 +101,10 @@ public static Transaction newChallenge(
         .addOperation(domainNameOperation)
         .addOperation(webAuthDomainOperation);
 
+    if (memo != null) {
+      builder.addMemo(memo);
```
JakeUrban, line 105: We need to ensure only memos of type ID are used here. Is there a subclass of memo for this type? Otherwise we should use a 64-bit integer as the parameter and create the memo object inside the function.

JakeUrban (changes requested):
We also need to validate that the memo is of type ID in the read challenge tx function.

src/main/java/org/stellar/sdk/Sep10Challenge.java
lijamie98: Makes sense. I had added the type check 
```java

      if (!(memo instanceof MemoId)) {
        throw new InvalidSep10ChallengeException("only memo type `Id` is supported");
      }

```

JakeUrban (approved):
This PR looks good for the `newChallenge()` endpoint, but I think we should also update `readChallengeTx()` to verify the memo attached, if present, is of type ID. Maybe that makes more sense in a 2nd PR, I'll leave it to you.

tamirms:
I agree with @JakeUrban , should we be validating the memo field during readChallengeTx?

lijamie98:
> I agree with @JakeUrban , should we be validating the memo field during readChallengeTx?

I had added memo validation to the readChallengeTx().

JakeUrban (approved):

PR #437, opened 2022-06-01
Modify CHANGELOG.md for 0.34.0

lijamie98:


tamirms (approved):

sreuland (approved):
@lijamie98 , it looks like you've got the release updates to CHANGELOG here and the build.gradle version=0.34.0 was merged to master on a prior PR, you could finish the GH 0.34.0 Release by merging this and then tag off master since they are same at this point before anything else goes in.  The [other process mentioned](https://docs.google.com/document/d/1vxGpbbTEsiT8WO_aprPrDWfBygSlsG-omUkNrMK9Ygg/edit?pli=1#) was to first create a `release-x.y.z` branch off master and then do the CHANGELOG and build.gradle updates in the  release branch and tag release off that and merge it back to master after release, but can effectively bypass that in this case as mentioned.

PR #440, opened 2022-06-30
bugfix: make `Sep10Challenge.verifyTransactionSignatures` resilient against account IDs that are not compliant with ed25519

### LINKED ISSUE #439: Fix crash in `Sep10Challenge.verifyChallengeTransactionThreshold` when the account id is not a valid ed25519 public key

[`0.34.0`](https://github.com/stellar/java-stellar-sdk/releases/tag/0.34.0)

### What did you do?

When calling `Sep10Challenge.verifyChallengeTransactionThreshold` with a list of signers that included the [`GA2T6GR7VXXXBETTERSAFETHANSORRYXXXPROTECTEDBYLOBSTRVAULT`](https://stellar.expert/explorer/public/account/GA2T6GR7VXXXBETTERSAFETHANSORRYXXXPROTECTEDBYLOBSTRVAULT) address (that's not compliant with ed25519), the code throwed an exception at line:

https://github.com/stellar/java-stellar-sdk/blob/d36410080c349c35dd9f6ec72f1e7a5202198d9b/src/main/java/org/stellar/sdk/Sep10Challenge.java#L555

I expected this account address to be handled graciously, either as a valid account or being ignored if there's no other way.

### What did you see instead?

The app threw a `RuntimeException("Public key is invalid")` at:
https://github.com/stellar/java-stellar-sdk/blob/8dcb082ff21bc8c6c55a2475bfb1652110a31f46/src/main/java/org/stellar/sdk/KeyPair.java#L116-L129

marcelosalloum:
### What
Make `Sep10Challenge.verifyTransactionSignatures` resilient against account IDs that are not compliant with ed25519. If a non-compliant account was provided in the list of signers, this account is simply ignored.

Accounts whose ID is not a valid ed25519 public key won't have a signing key, so they can't have signed the challenge transaction.

After the `Sep10Challenge.verifyTransactionSignatures` method finds the intersection between `account IDs that signed the transaction` ∩ `signers provided in the list`, it will return to the caller and the number of accounts and their weight will be verified accordingly.

### Why

Close #439.

### DIFF
--- a/src/main/java/org/stellar/sdk/Sep10Challenge.java
+++ b/src/main/java/org/stellar/sdk/Sep10Challenge.java
@@ -552,7 +552,14 @@ public class Sep10Challenge {
     }
 
     for (String signer : signers) {
-      KeyPair keyPair = KeyPair.fromAccountId(signer);
+      KeyPair keyPair;
+      try {
+        keyPair = KeyPair.fromAccountId(signer);
+      } catch (RuntimeException e) {
+        System.out.printf("Couldn't parse the signer \"%s\" as a valid Public Key.", signer);
+        e.printStackTrace();
+        continue;
+      }
       SignatureHint hint = keyPair.getSignatureHint();
 
       for (Signature signature : signatures.get(hint)) {
--- a/src/main/java/org/stellar/sdk/Sep10Challenge.java
+++ b/src/main/java/org/stellar/sdk/Sep10Challenge.java
@@ -556,8 +556,6 @@ public class Sep10Challenge {
       try {
         keyPair = KeyPair.fromAccountId(signer);
       } catch (RuntimeException e) {
-        System.out.printf("Couldn't parse the signer \"%s\" as a valid Public Key.", signer);
-        e.printStackTrace();
         continue;
       }
       SignatureHint hint = keyPair.getSignatureHint();
--- a/src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
+++ b/src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
@@ -1609,12 +1609,10 @@ public class Sep10ChallengeTest {
   }
 
   @Test
-  public void testVerifyChallengeTransactionThresholdValidServerAndMultipleClientKeyMeetingThresholdSomeUnusedAndOneNotEd25519Compliant() throws IOException, InvalidSep10ChallengeException {
+  public void testVerifyChallengeTransactionWithAccountIdNonCompliantWithEd25519() throws IOException, InvalidSep10ChallengeException {
     Network network = Network.TESTNET;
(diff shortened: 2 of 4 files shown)


marcelosalloum:
> Given that we `continue` on invalid signers, does this make it possible for a challenge to be "valid" despite being full of fake signers? I would expect at least one real additional signer to be required.

Maybe my description of the PR is not precise enough...

What I actually changed was the method `verifyTransactionSignatures(transaction, signers)`, that basically gets the intersection between `account IDs that signed the transaction` ∩ `signers provided in the list`.

After that, these signers will validated in the caller method, where this verification is performed:
[∩](https://github.com/stellar/java-stellar-sdk/blob/6d8a397969a9bca084ffb739de34744a6e12c258/src/main/java/org/stellar/sdk/Sep10Challenge.java#L435-L468)

If no signer is found or if the number of signers found is not sufficient, an exception will be thrown.

> Note: I'm updating the PR description to make it more precise.

src/test/java/org/stellar/sdk/Sep10ChallengeTest.java
```diff
@@ -1608,6 +1608,43 @@ public void testVerifyChallengeTransactionThresholdValidServerAndMultipleClientK
     assertEquals(new HashSet<String>(Arrays.asList(masterClient.getAccountId(), signerClient1.getAccountId())), signersFound);
   }
 
+  @Test
+  public void testVerifyChallengeTransactionThresholdValidServerAndMultipleClientKeyMeetingThresholdSomeUnusedAndOneNotEd25519Compliant() throws IOException, InvalidSep10ChallengeException {
```
sreuland, line 1612: now that's a test case name, good coverage to assert the change! maybe condense the name down to just the non-compliant aspect.

src/main/java/org/stellar/sdk/Sep10Challenge.java
```diff
@@ -552,7 +552,14 @@ private static Set<String> verifyTransactionSignatures(Transaction transaction,
     }
 
     for (String signer : signers) {
-      KeyPair keyPair = KeyPair.fromAccountId(signer);
+      KeyPair keyPair;
+      try {
+        keyPair = KeyPair.fromAccountId(signer);
+      } catch (RuntimeException e) {
+        System.out.printf("Couldn't parse the signer \"%s\" as a valid Public Key.", signer);
```
sreuland, line 559: this console output is non-standard for the sdk which by default does not generate that type of output or logger outputs. So, to maintain consistency, would remove this and the stacktrace prints. the callers of this method could check the returned set, and determine which `signers` are missing, then it could deduce something was wrong and log something from the app layer, etc.

sreuland (approved):
thanks for jumping in to make the required change!

Issue #446, opened 2022-07-21
0.35.0 'Build failed' error with the several 'Duplicate class' errors


let-it-snow:
### What version are you using?

Newest version **0.35.0**

### What did you do?
In the **0.35.0** version several particular libraries were updated (okhttp and others). These updates are causing the **‘Build failed’** error with several **‘Duplicate class’** errors.  This error occurred on the all projects using the latest java stellar SDK version. All projects are **Kotlin** projects.

Steps to reproduce:
 - Create a new Kotlin project in the Android Studio (I had used Kotlin **v1.7.10** or **v1.6.21**).
 - Apply Java Stellar SDK 0.35.0.
 - Try to build it.
 - Build failed with the several **'Duplicate class'** errors.
 
```
Duplicate class kotlin.ArrayIntrinsicsKt found in modules jetified-java-stellar-sdk-0.35.0 (com.github.stellar:java-stellar-sdk:0.35.0) and jetified-kotlin-stdlib-1.7.10 (org.jetbrains.kotlin:kotlin-stdlib:1.7.10)
Duplicate class kotlin.BuilderInference found in modules jetified-java-stellar-sdk-0.35.0 (com.github.stellar:java-stellar-sdk:0.35.0) and jetified-kotlin-stdlib-1.7.10 (org.jetbrains.kotlin:kotlin-stdlib:1.7.10)
............
Duplicate class org.intellij.lang.annotations.Subst found in modules jetified-annotations-16.0.1 (org.jetbrains:annotations:16.0.1) and jetified-java-stellar-sdk-0.35.0 (com.github.stellar:java-stellar-sdk:0.35.0)
Duplicate class org.jetbrains.annotations.Contract found in modules jetified-annotations-16.0.1 (org.jetbrains:annotations:16.0.1) and jetified-java-stellar-sdk-0.35.0 (com.github.stellar:java-stellar-sdk:0.35.0)
............

``` 

Do you have any fixes or suggestions on how to address this issue?

### What did you expect to see?
The Kotlin project with Java Stellar SDK 0.35.0 is built successfully.

### What did you see instead?
Incompatibility error.

Thanks.

tamirms:
@let-it-snow I have been able to reproduce the issue. I'm not sure yet how we can fix this on our side but I think you can fix this in your Kotlin project by adding the following clause (right below the `dependencies {}` block) to your build.gradle file:

```
configurations.implementation {
    exclude group: 'org.jetbrains.kotlin'
[...]

PR #448, opened 2022-07-25
Shade kotlin dependencies

tamirms:
Close https://github.com/stellar/java-stellar-sdk/issues/446

In the 0.35.0 release we upgraded the com.squareup.okhttp3:okhttp dependency from v3 to v4. Bumping that dependency pulled in a dependency on kotlin. This causes "Duplicate Class" errors when using the java stellar sdk on kotlin projects. The fix for this issue is to properly shade all the kotlin dependencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ As this project is pre 1.0, breaking changes may happen for minor version bumps.
 
 ## Pending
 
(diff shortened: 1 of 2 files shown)


lijamie98 (approved):
LGTM.

sreuland (approved):

PR #457, opened 2022-12-02
fix okhttp3 class loader issue on android platform

### LINKED ISSUE #450: 0.36.0 Crash on Android <=9 -> okhttp3 ExceptionInInitializerError

Newest version **0.36.0**

### What did you do?

Create the `org.stellar.sdk.Server("https://horizon.stellar.co/")` object in any place of the project -> You will receive the crash: `java.lang.ExceptionInInitializerError` on the **Android version <=9**

```
StandardAndroidSocketAdapter.kt line 61
....
java.lang.ExceptionInInitializerError
        at shadow.okhttp3.OkHttpClient.<init>(OkHttpClient.kt:237)
        at shadow.okhttp3.OkHttpClient$Builder.build(OkHttpClient.kt:1069)
        at org.stellar.sdk.Server.<init>(Server.java:58)
```

Steps to reproduce:

- Create a new Kotlin project in the Android Studio.
- Apply Java Stellar SDK 0.36.0.
- Create `org.stellar.sdk.Server("https://horizon.stellar.co/") `object in any place.
- Run the project on the **Android <= 9** and immediately receive the `ExceptionInInitializerError`.

No any crashes.

### What did you see instead?
Crash on the **Android <=9**.

Thanks.

sreuland:
I was able to replicate this problem in Android studio using emulator set at SDK version 28( android 'pie'):

The underlying issue was due to the shadowed jar, which rewrote an unintended 'com.android.org.conscrypt' hardcoded reference in the okhttp3 platform initialization which uses that as default to initialize networking from android:

https://github.com/square/okhttp/blob/parent-4.10.0/okhttp/src/main/kotlin/okhttp3/internal/platform/android/StandardAndroidSocketAdapter.kt#L61

it had a cycle when in error, trying to reference the platform instance while it was inside the platform init routine.

to avoid the problem, I tweaked the shadow config to ignore `com.android.org.conscrypt` ,which stops it from trying to rewrite any refs to that in the .class files in the shadowed jar. 

after this change, I was able to verify the fix by running the same prior failing example with `org.stellar.sdk.Server("https://horizon.stellar.com/")` and `OkHttpClient()` which now worked.

Closes #450

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,7 +39,11 @@ shadowJar {
     duplicatesStrategy DuplicatesStrategy.EXCLUDE
     archiveClassifier.set('')
     archiveFileName.set('stellar-sdk.jar')
-    relocate 'com.','shadow.com.'
+    relocate('com.', 'shadow.com.') {
+      // okhttp hardcodes android platform class loading to this package, shadowing would attempt to rewrite the hardcode
+      // to be 'shadow.com.android.org.conscrypt' which we don't want to happen.
+      exclude 'com.android.org.conscrypt'
+    }
     relocate 'net.','shadow.net.'
     relocate 'javax.annotation', 'shadow.javax.annotation'
     relocate 'org.apache','shadow.org.apache'


marcelosalloum (approved):
LGTM!

tsachiherman (approved):

tsachiherman:
It would be great if we had a unit test that reproduces this issue - but I don't know how feasible it would be to implement such a test.

sreuland:
> It would be great if we had a unit test that reproduces this issue - but I don't know how feasible it would be to implement such a test.

yes, created ticket to track - https://github.com/stellar/java-stellar-sdk/issues/458

PR #474, opened 2023-04-19
Fix NPE

Ifropc:


leighmcculloch (approved):
👏🏻 Nice catch.

src/test/java/org/stellar/sdk/KeyPairTest.java
```diff
@@ -120,4 +120,11 @@ public void testSignPayloadSignerLessThanHint() {
     // the hint could only be derived off of 3 bytes from payload
     Assert.assertArrayEquals(sig.getHint().getSignatureHint(), new byte[]{(byte)(255), 64, 7, 55});
   }
+
+  @Test
+  public void testPublicEqual() {
+    KeyPair keypair = KeyPair.fromAccountId("GDEAOZWTVHQZGGJY6KG4NAGJQ6DXATXAJO3AMW7C4IXLKMPWWB4FDNFZ");
+    KeyPair keypairCopy = KeyPair.fromAccountId("GDEAOZWTVHQZGGJY6KG4NAGJQ6DXATXAJO3AMW7C4IXLKMPWWB4FDNFZ");
+    Assert.assertEquals(keypairCopy, keypair);
+  }
```
leighmcculloch, line 129: Thanks for adding a test for the case being fixed. I don't think there's a test for equals in the cases that one keypair is private and one is public, or that both are private. If those tests don't exist elsewhere that I'm missing, could we add them too?

leighmcculloch (approved):
👏🏻 Nice catch.

src/test/java/org/stellar/sdk/KeyPairTest.java
Ifropc: Done

leighmcculloch (approved):
🎉

Issue #476, opened 2023-04-27
CRASH when handling 429s


marcelosalloum:
### What version are you using?

0.34.1

### What did you do?

Hit Horizon too many times too fast.

### What did you expect to see?

An understandable error.

### What did you see instead?

A NullPointerException error:

```java
21:31:40.507 ERROR - o.s.a.u.Log - java.lang.NumberFormatException: null
	at java.base/java.lang.Integer.parseInt(Integer.java:614)
	at java.base/java.lang.Integer.parseInt(Integer.java:770)
	at org.stellar.sdk.requests.ResponseHandler.handleResponse(ResponseHandler.java:31)
	at org.stellar.sdk.requests.AccountsRequestBuilder.account(AccountsRequestBuilder.java:40)
	at org.stellar.sdk.requests.AccountsRequestBuilder.account(AccountsRequestBuilder.java:51)
	at org.stellar.anchor.sep10.Sep10Service.validateChallenge(Sep10Service.java:234)
	at org.stellar.anchor.sep10.Sep10Service.validateChallenge(Sep10Service.java:191)
	at org.stellar.anchor.platform.controller.sep.Sep10Controller.validateChallenge(Sep10Controller.java:83)
	at org.stellar.anchor.platform.controller.sep.Sep10Controller.validateChallenge(Sep10Controller.java:71)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.springframework.web.method.support.InvocableHandlerMethod.doInvoke(InvocableHandlerMethod.java:205)
	at org.springframework.web.method.support.InvocableHandlerMethod.invokeForRequest(InvocableHandlerMethod.java:150)
	at org.springframework.web.servlet.mvc.method.annotation.ServletInvocableHandlerMethod.invokeAndHandle(ServletInvocableHandlerMethod.java:117)
	at org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.invokeHandlerMethod(RequestMappingHandlerAdapter.java:895)
[...]

PR #477, opened 2023-04-28
[ANCHOR-264] Fix Null Point Exception when reaching request max retries

paulormnas:
Fix Issue [476](https://github.com/stellar/java-stellar-sdk/issues/476) of Null Point Exception on reaching a max retries of Account Builder Request.

### DIFF
--- a/src/main/java/org/stellar/sdk/requests/TooManyRequestsException.java
+++ b/src/main/java/org/stellar/sdk/requests/TooManyRequestsException.java
@@ -5,17 +5,7 @@ package org.stellar.sdk.requests;
  * @see <a href="https://developers.stellar.org/api/introduction/rate-limiting/" target="_blank">Rate Limiting</a>
  */
 public class TooManyRequestsException extends RuntimeException {
-  private int retryAfter;
-
-  public TooManyRequestsException(int retryAfter) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/stellar/sdk/requests/TooManyRequestsException.java
```diff
@@ -5,17 +5,7 @@
  * @see <a href="https://developers.stellar.org/api/introduction/rate-limiting/" target="_blank">Rate Limiting</a>
  */
 public class TooManyRequestsException extends RuntimeException {
-  private int retryAfter;
-
-  public TooManyRequestsException(int retryAfter) {
+  public TooManyRequestsException() {
```
Ifropc, line 8: retyAfter variable can still be used by the outside developer.
You should either give some reasonable default, or make retryAfter a nullable Integer

Ifropc (changes requested):

src/main/java/org/stellar/sdk/requests/TooManyRequestsException.java
paulormnas: I'm closing this PR. There's already another PR opened that just need to merge

PR #480, opened 2023-07-08
Format code

overcat:
This PR mainly adds a code formatting tool for better collaboration. The second commit formats the code. If you need to review the code, please check the changes in the first and third commits.


overcat:
> I think `scripts/pre-commit.sh` should be executable (e.g. run `chmod +x scripts/pre-commit.sh`). Otherwise, this PR looks good!

Before, I had set the file as an executable in build.gradle. Now, I have already marked the original file as executable, see b838e16911fecace14b0fa829e7ec8d3d74ba61d

Issue #486, opened 2023-07-20
Add support for BumpFootprintExpirationOperation and RestoreFootprintOperation Responses.


overcat:
### What problem does your feature solve?

BumpFootprintExpirationOperation and RestoreFootprintOperation are newly introduced operations in the Protocol Soroban, and we need to add corresponding support for them.

### What would you like to see?
- Allow users to construct transactions that include these two operations.
- The server needs to be able to parse these two operations correctly. ([protocol](https://github.com/stellar/go/blob/soroban-xdr-next/protocols/horizon/operations/main.go#L369-L381))

### What alternatives are there?
None

Estimated working hours: 6

sreuland:
I updated the title to narrow the scope to just supporting response models. The TransactionBuilder and TxSub aspects of these ops will need a new ticket created like 'Support TxSub for Invoke/Bump/Restore', and that will require soroban rpc integration and new flows for assembling transactions with preflight response data for setting tx footprint, fees. the js sdk, js-soroban-client, has gone through this enhancement for submitting tx's with these ops and preflighting, which can be used as a reference - https://github.com/stellar/js-soroban-client/pull/108

overcat:
Implemented in #491.

Issue #485, opened 2023-07-20
Add support for Soroban RPC server.


overcat:
### What problem does your feature solve?
Enable users to invoke Soroban RPC services through this SDK

### What would you like to see?
We need to add the following functions:
- getAccount
- getHealth
- getNetwork
- getContractData
- getLedgerEntries
- getTransaction
- getEvents
- getLatestLedger
- simulateTransaction
- prepareTransaction
- sendTransaction

### What alternatives are there?
None

Estimated working hours: 24

sreuland:
can refer to the javascript sdk, [server.ts](https://github.com/stellar/js-soroban-client/blob/main/src/server.ts), for reference example of sdk method signatures for wrapping the soroban rpc endpoints.

overcat:
Implemented in #492.

PR #491, opened 2023-07-21
Add support for BumpFootprintExpirationOperation and RestoreFootprintOperation

overcat:
close https://github.com/stellar/java-stellar-sdk/issues/486


overcat:
Hi @sreuland, the PR is ready, could you please review and merge it?

src/main/java/org/stellar/sdk/BumpFootprintExpirationOperation.java
```diff
@@ -0,0 +1,71 @@
...
+import lombok.Value;
+import lombok.experimental.SuperBuilder;
+import org.stellar.sdk.xdr.BumpFootprintExpirationOp;
+import org.stellar.sdk.xdr.ExtensionPoint;
+import org.stellar.sdk.xdr.OperationType;
+import org.stellar.sdk.xdr.Uint32;
+
+/**
+ * Represents <a
+ * href="https://developers.stellar.org/docs/fundamentals-and-concepts/list-of-operations#bump-footprint-expiration"
```
sreuland, line 14: this docs link isn't valid yet, the public docs haven't been updated for the new soroban ops yet. Can you update these references to point at the go struct defs for the json model instead, it's located in the `soroban-xdr-next` branch of [stellar/go/protocols/horizon/main](https://github.com/stellar/go/blob/soroban-xdr-next/protocols/horizon/operations/main.go#L376-L381)

src/test/java/org/stellar/sdk/BumpFootprintExpirationOperationTest.java
```diff
@@ -0,0 +1,111 @@
...
+  public void testBuilderWithSource() {
+    String source = "GASOCNHNNLYFNMDJYQ3XFMI7BYHIOCFW3GJEOWRPEGK2TDPGTG2E5EDW";
+    BumpFootprintExpirationOperation op =
+        BumpFootprintExpirationOperation.builder()
+            .ledgersToExpire(123)
+            .sourceAccount(source)
+            .build();
+    assertEquals(Integer.valueOf(123), op.getLedgersToExpire());
+    assertEquals(source, op.getSourceAccount());
+    String expectXdr = "AAAAAQAAAAAk4TTtavBWsGnEN3KxHw4Ohwi22ZJHWi8hlamN5pm0TgAAABkAAAAAAAAAew==";
```
sreuland, line 20: are there other tests in this repo that hardcode expected xdr, that may not be a good pattern to follow, can make the test fragile as the xdr may change, expand over time. Instead, can do a full round trip with marshaling such as `secondOp = Operation.fromXdr(op.toXdrBase64())` and then `assertEquals(secondOp, op)`

overcat: I tend to keep it. Our XDR module code lacks sufficient testing, and these XDR values are generated by other SDKs (Python, JS), which can serve as cross-validation. Additionally, changes in XDR definitions do not occur frequently, so even if we update it with some effort, it is acceptable.

In the previous code, there were also some validations done in this way. https://github.com/stellar/java-stellar-sdk/blob/master/src/test/java/org/stellar/sdk/OperationTest.java#L167

src/main/java/org/stellar/sdk/BumpFootprintExpirationOperation.java
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * <p>Bump the expiration of a footprint (read and written ledger keys).
+ *
+ * @see <a href="https://developers.stellar.org/docs/fundamentals-and-concepts/list-of-operations"
+ *     target="_blank">List of Operations</a>
+ */
+@EqualsAndHashCode(callSuper = true)
+@SuperBuilder(toBuilder = true)
+@Value
+public class BumpFootprintExpirationOperation extends Operation {
```
sreuland, line 25: can you add TODO's on these that state invoke, bump, restore soroban op classes in this package still require more infrastructure before they can be used in TransactionBuilder and submission, they can't be added as an Operation in a standard 'classic' tx. 

Since they have to go through preflight with rpc first, to define footprint, fees and then build out a single operation tx with all that. the js sdk, js-soroban-client, [has added this functionality for tx sub of soroban ops](https://github.com/stellar/js-soroban-client/pull/108) which can be used as a reference later on for same effort here. I think we should add a new ticket to capture this effort such as 'Submit Soroban Tx operations' which will depend on #485 

I [updated and commented on #486](https://github.com/stellar/java-stellar-sdk/issues/486#issuecomment-1652819596) to reflect the narrowing of scope for this PR to be primarily for the response object marshaling.

src/main/java/org/stellar/sdk/Operation.java
```diff
@@ -209,6 +209,12 @@ public static Operation fromXdr(
       case INVOKE_HOST_FUNCTION:
         operation = InvokeHostFunctionOperation.fromXdr(body.getInvokeHostFunctionOp());
         break;
+      case BUMP_FOOTPRINT_EXPIRATION:
+        operation = BumpFootprintExpirationOperation.fromXdr(body.getBumpFootprintExpirationOp());
```
sreuland, line 213: I think we should throw a NotSupportException here for invoke/bump/restore for now and add a TODO referring to https://github.com/stellar/java-stellar-sdk/issues/495

src/main/java/org/stellar/sdk/responses/operations/BumpFootprintExpirationOperationResponse.java
```diff
@@ -0,0 +1,24 @@
...
+ * Represents BumpFootprintExpiration operation response.
+ *
+ * @see <a href="https://developers.stellar.org/api/horizon/resources/operations"
+ *     target="_blank">Operation documentation</a>
+ * @see org.stellar.sdk.requests.OperationsRequestBuilder
+ * @see org.stellar.sdk.Server#operations()
+ */
+public class BumpFootprintExpirationOperationResponse extends OperationResponse {
+  @SerializedName("ledgers_to_expire")
+  private final Long ledgers_to_expire;
```
sreuland, line 18: I don't think snake casing is used as a code style, this should probably be mentioned in #493, using camel case insted such as `ledgersToExpire` is predominant format.

src/main/java/org/stellar/sdk/BumpFootprintExpirationOperation.java
overcat: Fixed in 3eb0d1a63707653df7ee111b3682f729127c6197, I also updated the links in the code related to InvokeHostFunction.

src/main/java/org/stellar/sdk/responses/operations/BumpFootprintExpirationOperationResponse.java
sreuland: could probably make this a lombok @Value class also, and not implement any constructor or getter/setters

src/main/java/org/stellar/sdk/Operation.java
overcat: We will not merge into the master branch now, but into the soroban branch instead. We can assume that all necessary work will be completed before merging the code into the main branch. Therefore, I prefer to track these TODO items in the issue rather than in the code itself. What do you think?

src/main/java/org/stellar/sdk/responses/operations/BumpFootprintExpirationOperationResponse.java
overcat: Sorry, my mistake, fixed in 758746597f480d1df08d55ec0b279dfc61867a89

src/test/java/org/stellar/sdk/BumpFootprintExpirationOperationTest.java
sreuland: yes, I have seen the usage of hardcoded xdr for serialization tests, not suggesting to change either, but that is more for scope in the xdrgen library to verify. This unit test's responsibility is to validate the code from this sdk builds a valid java object graph for BumpFootprint, and it has asserted that, it doesn't seem like the extra assertions on serialized xdr are related to code in this repo. But, I am relatively newer to this repo, and would be great if @tamirms wants to jump in with any thoughts on this also.

src/main/java/org/stellar/sdk/Operation.java
sreuland: yes, the feature branch and no-one else is developing on it or using it incrementally allows for that and will leave to your discretion. On it's own, It looks like it's complete but reality is not, so, adding the NotSupportedFunction here in code and a unit test that asserts that as outcome for now is just a way to just make this clear right in the code that it is currently not complete.

overcat: OK, actually I have been working on this related task in the soroban-server branch, so let's keep it that way for now.

src/main/java/org/stellar/sdk/BumpFootprintExpirationOperation.java
overcat: I have decided not to modify it for now, see https://github.com/stellar/java-stellar-sdk/pull/491#discussion_r1275692913

sreuland: ok, I think understand, you will leave the classes in this pr, and once it merges to [soroban](https://github.com/stellar/java-stellar-sdk/tree/soroban), then the wip there in https://github.com/stellar/java-stellar-sdk/pull/492 will refer to them?

src/main/java/org/stellar/sdk/Operation.java
sreuland: ok, understood, that is https://github.com/stellar/java-stellar-sdk/pull/492 correct, it will be first to reference this aspect?

sreuland (approved):
@overcat , this looks good, responded on comments, is it ready to merge?

overcat:
> @overcat , this looks good, responded on comments, is it ready to merge?

Yes, please merge this PR first. #492 depends on this PR.

PR #481, opened 2023-07-13
Add support for InvokeHostFunctionOperation & add `Address` class.

overcat:
## Changelog:
- Add support for InvokeHostFunctionOperation. (Closes #472)
- Add support for parsing InvokeHostFunctionOperationResponse. 
- Add `Address` class, it represents a single address in the Stellar network. 
- Updated xdr definitions for Stellar Soroban & fix the issues caused by the XDR update.

## Note:
This PR depends on stellar/xdrgen#159, stellar/xdrgen#160 and stellar/xdrgen#161, let's also merge these PRs.

PR #492, opened 2023-07-21
[WIP] Added support for calling Soroban RPC server.

### LINKED ISSUE #485: Add support for Soroban RPC server.

overcat:
close #485


sreuland:
make sure to change merge target on this to 'soroban' branch

sreuland:
assume this depends on https://github.com/stellar/java-stellar-sdk/pull/481 to merge first as it will bring in all the new xdr definitions?

overcat:
> assume this depends on #481 to merge first as it will bring in all the new xdr definitions?

Yes, let's merge #481 first.

src/main/java/org/stellar/sdk/LedgerEntryNotFoundException.java
```diff
@@ -0,0 +1,14 @@
+package org.stellar.sdk;
+
+public class LedgerEntryNotFoundException extends Exception {
```
sreuland, line 3: maybe a style guideline to add, for any public class at minimum, add a javadoc for the class level, and would be nice to have javadoc on any public methods.

src/main/java/org/stellar/sdk/SorobanServer.java
```diff
@@ -0,0 +1,344 @@
...
+
+/**
+ * Main class used to connect to the Soroban-RPC instance and exposes an interface for requests to
+ * that instance.
+ */
+public class SorobanServer implements Closeable {
+  private static final int SUBMIT_TRANSACTION_TIMEOUT = 60; // seconds
+  private static final int CONNECT_TIMEOUT = 10; // seconds
+  private final HttpUrl serverURI;
+  private final OkHttpClient httpClient;
```
sreuland, line 61: maybe this could be an opportunity to replace OkHttpClient because it's really bloated and it pulls in kotlin as a dependency even though it's not used here at all. 

If you want to evaluate http clients a bit, the other alternatives are apache HttpClient or bump the java jdk level of project to 11 and then can use the jdk's built-in `java.net.http.HttpClient`

src/main/java/org/stellar/sdk/SorobanServer.java
```diff
@@ -0,0 +1,344 @@
...
+            .retryOnConnectionFailure(true)
+            .build());
+  }
+
+  public SorobanServer(String serverURI, OkHttpClient httpClient) {
+    this.serverURI = HttpUrl.parse(serverURI);
+    this.httpClient = httpClient;
+  }
+
+  public TransactionBuilderAccount getAccount(String accountId)
```
sreuland, line 85: would be helpful to javadoc the public methods, both for readability of source and also for client usage.

src/main/java/org/stellar/sdk/SorobanServer.java
```diff
@@ -0,0 +1,344 @@
...
+      throw new LedgerEntryNotFoundException(ledgerKeyToXdrBase64(ledgerKey));
+    }
+
+    return entries.get(0);
+  }
+
+  public GetLedgerEntriesResponse getLedgerEntries(Collection<LedgerKey> keys) throws IOException {
+    List<String> xdrKeys =
+        keys.stream().map(SorobanServer::ledgerKeyToXdrBase64).collect(Collectors.toList());
+    GetLedgerEntriesRequest params = new GetLedgerEntriesRequest(xdrKeys);
```
sreuland, line 157: is an empty `keys` list a valid  request and/or should this check for empty list on `xdrKeys`, to avoid sending to rpc in any case?

src/main/java/org/stellar/sdk/SorobanServer.java
```diff
@@ -0,0 +1,344 @@
...
+      throw new LedgerEntryNotFoundException(ledgerKeyToXdrBase64(ledgerKey));
+    }
+    LedgerEntry.LedgerEntryData ledgerEntryData =
+        ledgerEntryDataFromXdrBase64(entries.get(0).getXdr());
+    long sequence = ledgerEntryData.getAccount().getSeqNum().getSequenceNumber().getInt64();
+    return new Account(accountId, sequence);
+  }
+
+  public GetHealthResponse getHealth() throws IOException {
+    return this.<Void, GetHealthResponse>sendRequest(
```
sreuland, line 110: can remove the `this` references, that scope is implied when any non-static method invokes another non-static method of class.

src/main/java/org/stellar/sdk/requests/sorobanrpc/SorobanRpcErrorResponse.java
```diff
@@ -0,0 +1,19 @@
+package org.stellar.sdk.requests.sorobanrpc;
+
+import lombok.Getter;
+
+@Getter
+public class SorobanRpcErrorResponse extends RuntimeException {
```
sreuland, line 6: i expected this to be checked exception, forcing callers to handle it, what was the reason for coercing this to unchecked with `extends RuntimeException`?

sreuland: hmm, after more review, I don't think we want to model an rpc error response as an exception, rather it is conveyed as an attribute in the SorobanRpcResponse instead, thinking this SorobanRpcErrorResponse can be removed?

src/main/java/org/stellar/sdk/SorobanServer.java
```diff
@@ -0,0 +1,344 @@
...
+    ResponseHandler<SorobanRpcResponse<R>> responseHandler = new ResponseHandler<>(responseType);
+    SorobanRpcRequest<T> sorobanRpcRequest = new SorobanRpcRequest<>(requestId, method, params);
+    MediaType mediaType = MediaType.parse("application/json");
+    RequestBody requestBody =
+        RequestBody.create(gson.toJson(sorobanRpcRequest).getBytes(), mediaType);
+
+    Request request = new Request.Builder().url(this.serverURI).post(requestBody).build();
+    try (Response response = this.httpClient.newCall(request).execute()) {
+      SorobanRpcResponse<R> sorobanRpcResponse = responseHandler.handleResponse(response);
+      if (sorobanRpcResponse.getError() != null) {
```
sreuland, line 274: thinking about this, not sure that a successful http response with json-rpc error in the response payload should be elevated to an exception, instead, just return the `sorobanRpcResponse` in all cases if you get it, as it has result and error attributes, that way clients only have to catch the IOException as a fatal type problem, otherwise they will inspect the response [result/error](https://github.com/stellar/java-stellar-sdk/pull/492/files#diff-db0499c80b2c87816a70ab620b64dea366e1b5d76c2fec3593c73db248d13a25R22) for further application meaning. I [commented similar on SorobanRpcErrorResponse](https://github.com/stellar/java-stellar-sdk/pull/492/files#r1276796480)

src/main/java/org/stellar/sdk/requests/sorobanrpc/SorobanRpcRequest.java
```diff
@@ -0,0 +1,21 @@
+package org.stellar.sdk.requests.sorobanrpc;
+
+import com.google.gson.annotations.SerializedName;
+import lombok.RequiredArgsConstructor;
+import lombok.Value;
+
+@RequiredArgsConstructor
+@Value
+public class SorobanRpcRequest<T> {
```
sreuland, line 9: nice, doing the json-rpc protocol manually on top of just an http client here looks good, rather than trying to use an external lib that does the same but would bloat the dependencies, etc.

src/main/java/org/stellar/sdk/responses/sorobanrpc/SendTransactionResponse.java
```diff
@@ -0,0 +1,38 @@
+package org.stellar.sdk.responses.sorobanrpc;
+
+import com.google.gson.annotations.SerializedName;
+import lombok.AllArgsConstructor;
+import lombok.Value;
+
+@AllArgsConstructor
+@Value
+public class SendTransactionResponse {
+  @SerializedName("status")
```
sreuland, line 10: can remove the @SerializedName annotations when it matches the java class member name, gson will by default serialize to that name.

src/main/java/org/stellar/sdk/SorobanServer.java
sreuland: oh, skip taking that on in here, we should just carve out changing the jdk level and updating library deps to a separate 'tech debt' ticket so it doesn't conflate the effort here, thanks!

https://github.com/stellar/java-stellar-sdk/issues/496

cc @mollykarcher , @tamirms

overcat:
Hi @sreuland, this PR is still in WIP status. I will let you know when it is ready for review. Thank you!

src/main/java/org/stellar/sdk/SorobanServer.java
overcat: Yes, personally I prefer returning the entire `SorobanRpcResponse`, but after reading the code of other libraries (including js-soroban-sdk), I found that they directly return RpcResponse.result.

overcat: I have a question, is it necessary to expose the "id" field in jsonrpc to users? In other words, when making a jsonrpc request via http, is the "id" field almost useless?

overcat:
This PR is almost complete, of course there are still a few documents and test cases waiting to be added, welcome to leave comments.

overcat:
Click to view an example of creating and submitting a Soroban transaction.
<details>
  <summary>Click me.</summary>


```java
package org.stellar.sdk;

import java.io.IOException;
import org.junit.Test;
import org.stellar.sdk.requests.sorobanrpc.SorobanRpcErrorResponse;
import org.stellar.sdk.responses.sorobanrpc.SendTransactionResponse;
import org.stellar.sdk.xdr.HostFunction;
import org.stellar.sdk.xdr.HostFunctionType;
import org.stellar.sdk.xdr.SCSymbol;
import org.stellar.sdk.xdr.SCVal;
import org.stellar.sdk.xdr.SCValType;
import org.stellar.sdk.xdr.SCVec;
import org.stellar.sdk.xdr.Uint32;
import org.stellar.sdk.xdr.XdrString;

public class DemoTest {
  @Test
  public void demo()
      throws IOException,
          SorobanRpcErrorResponse,
          AccountNotFoundException,
          PrepareTransactionException {
    SorobanServer sorobanServer = new SorobanServer("http://100.83.15.43:8000/soroban/rpc");

    String contractId = "CDCYWK73YTYFJZZSJ5V7EDFNHYBG4QN3VUNG2IGD27KJDDPNCZKBCBXK";
    KeyPair txSubmitterKp =
        KeyPair.fromSecretSeed("SAAPYAPTTRZMCUZFPG3G66V4ZMHTK4TWA6NS7U4F7Z3IMUD52EK4DDEV");
    KeyPair opInvokerKp =
        KeyPair.fromSecretSeed("SAEZSI6DY7AXJFIYA4PM6SIBNEYYXIEM2MSOTHFGKHDW32MBQ7KVO6EN");

    TransactionBuilderAccount source = sorobanServer.getAccount(txSubmitterKp.getAccountId());

    Transaction transaction =
        new TransactionBuilder(AccountConverter.enableMuxed(), source, Network.STANDALONE)
            .setBaseFee(50000)
            .addPreconditions(
                TransactionPreconditions.builder().timeBounds(new TimeBounds(0, 0)).build())
            .addOperation(
                InvokeHostFunctionOperation.builder()
                    .sourceAccount(opInvokerKp.getAccountId())
                    .hostFunction(
                        new HostFunction.Builder()
                            .discriminant(HostFunctionType.HOST_FUNCTION_TYPE_INVOKE_CONTRACT)
                            .invokeContract(
                                new SCVec(
                                    new SCVal[] {
                                      new Address(contractId).toSCVal(),
                                      new SCVal.Builder()
                                          .discriminant(SCValType.SCV_SYMBOL)
                                          .sym(new SCSymbol(new XdrString("increment")))
                                          .build(),
                                      new Address(opInvokerKp.getAccountId()).toSCVal(),
                                      new SCVal.Builder()
                                          .discriminant(SCValType.SCV_U32)
                                          .u32(new Uint32(10))
                                          .build()
                                    }))
                            .build())
                    .build())
            .build();

    Transaction newTx = sorobanServer.prepareTransaction(transaction);

    newTx.sign(txSubmitterKp);
    newTx.sign(opInvokerKp);

    SendTransactionResponse resp = sorobanServer.sendTransaction(newTx);
    System.out.println(resp);
  }
}
```
</details>

src/main/java/org/stellar/sdk/SorobanServer.java
sreuland: just in case, I'd wire it in for allowing it to be set by caller in sdk, but accepting null,etc so that a default hardcoded id can be applied, just so we don't get stuck.

overcat: If that's the case, I will still stick to the current design, which is to throw a `SorobanRpcErrorResponse` exception if the response contains an error.

However, I will wrap the returned result, for example:
```java
@Value
@AllArgsConstructor
class WrappedResponse<T> {
  String id;
  T result;
}

public WrappedResponse<TransactionBuilderAccount> getAccount(String accountId)
```
Because interfaces like `getAccount` and `prepareTransaction` call RPC interfaces internally, but these interfaces do not directly return the data returned by Soroban-RPC to the user. This modification helps users obtain the ID.

src/main/java/org/stellar/sdk/responses/sorobanrpc/SendTransactionResponse.java
overcat: Fixed.

src/main/java/org/stellar/sdk/SorobanServer.java
overcat: After careful consideration, I suggest sticking with the current design. If there is a need to expose the `id` in the future, we can add it without introducing any breaking changes..

src/main/java/org/stellar/sdk/LedgerEntryNotFoundException.java
overcat: Fixed.

src/main/java/org/stellar/sdk/SorobanServer.java
overcat: Fixed.

src/main/java/org/stellar/sdk/SorobanServer.java
overcat: Fixed.

src/main/java/org/stellar/sdk/SorobanServer.java
overcat: Yes, but I think adding "this" is also very common. If you insist on removing it, I can do that.

src/main/java/org/stellar/sdk/requests/sorobanrpc/SorobanRpcErrorResponse.java
overcat: After the discussion in https://github.com/stellar/java-stellar-sdk/pull/492#discussion_r1278707543, I think it would be better to directly return the result instead of SorobanRpcErrorResponse. What do you think?

overcat:
Hi @sreuland, this PR is ready.

src/main/java/org/stellar/sdk/SorobanServer.java
sreuland: it's not common to use `this` when the variable scope doesn't require it, one example of usage is when a method level variable shadows a class variable, but the usage here in this scope is not required and results in some code bloat.

Issue #487, opened 2023-07-20
Add abstractions for converting between ScVal and native types.


overcat:
### What problem does your feature solve?
Currently, constructing ScVal directly is a rather painful task, so we need to add some aliases to make it easier for users to construct ScVal.

### What would you like to see?
I currently plan to refer to the implementation in the Python SDK. See https://github.com/StellarCN/py-stellar-base/tree/soroban/stellar_sdk/soroban/types

### What alternatives are there?
None

Estimated working hours: 16

overcat:
Implemented in #500.

PR #498, opened 2023-08-02
Correct the data type of certain fields to store the expected design values.

### LINKED ISSUE #494: Uint32 should not be represented using Integer.
c258f0d5cf4b734e2bed4ccfbcf3c27dcee0dcb4

### What did you do?
https://github.com/stellar/java-stellar-sdk/blob/c258f0d5cf4b734e2bed4ccfbcf3c27dcee0dcb4/src/main/java/org/stellar/sdk/xdr/Uint32.java#L15

Uint32 represents an unsigned 32-bit integer, which has a range of 0 to 2^32-1.
Integer in Java represents a signed 32-bit integer, which has a range of -2^31 to 2^31-1.

We should use Long instead of Integer.

### What did you see instead?
N/A

### LINKED ISSUE #499: Correct the data type of certain fields to store the expected design values.
latest code.

### What did you do?
Check #494 first.

Currently, some fields are using data types that are insufficient to hold the expected design values. As [an example](https://horizon-testnet.stellar.org/transactions/02eba0bdc423ef8ea87f51a4fb3f3f8644cac8165f2be744e87863fef87d07fb), memo.id = 18446744073709551615, and in this SDK, memo id is stored using `Long`. If this SDK is used to parse this transaction, the correct result cannot be obtained. We should modify it to be stored using `BigInteger`.

The following fields should be corrected.

| field                       | before         | after             |
|-----------------------------|----------------|----------------|
| LedgerBounds.minLedger       | int            | long           |
| LedgerBounds.maxLedger       | int            | long           |
| MemoId.id                    | long           | BigInteger     |
| TimeBounds.minTime           | long           | BigInteger     |
| TimeBounds.maxTime           | long           | BigInteger     |
| TransactionBuilder.baseFee   | int            | long           |
| TransactionPreconditions.TIMEOUT_INFINITE | long | BigInteger |
| TransactionPreconditions.minSeqAge | Long | BigInteger |
| TransactionPreconditions.minSeqLedgerGap | int | long |

These field types have been corrected.

### What did you see instead?
N/A

overcat:
Closes #494, Closes #499

#487 depends on this PR.

xdrgen-related PR is in https://github.com/stellar/xdrgen/pull/163

This PR includes breaking changes.

The types of the following fields have changed.

| field                       | before         | after             |
|-----------------------------|----------------|----------------|
| LedgerBounds.minLedger       | int            | long           |
| LedgerBounds.maxLedger       | int            | long           |
| MemoId.id                    | long           | BigInteger     |
| TimeBounds.minTime           | long           | BigInteger     |
| TimeBounds.maxTime           | long           | BigInteger     |
| TransactionBuilder.baseFee   | int            | long           |
| TransactionPreconditions.TIMEOUT_INFINITE | long | BigInteger |
| TransactionPreconditions.minSeqAge | Long | BigInteger |
| TransactionPreconditions.minSeqLedgerGap | int | long |


overcat:
I have added some TODO items and I am considering whether to complete them in this PR, but it would bring breaking changes.

sreuland:
@overcat , I'm starting review, I've looked at thehttps://github.com/stellar/xdrgen/pull/163, and it looks ok but I asked for other reviewers on that one as I don't have much experience there.

sreuland:
one small suggestion, could you update the docs for xdr generation in  [readme.md ](https://github.com/overcat/java-stellar-sdk/blob/fix-xdr-unsigned/readme.md?plain=1#L42-L46) to reflect the latest command needed to run, I'm not sure it's current for the new soroban `.x` files ?

overcat:
This command is already up to date, but if you want to generate the xdr classes in this PR, you need to merge the following [four PRs](https://github.com/stellar/xdrgen/pulls?q=is%3Aopen+is%3Apr+author%3A%40me+java).

src/main/java/org/stellar/sdk/xdr/XdrUnsignedHyperInteger.java
sreuland, line 1: just to confirm, you built this file first with xdrgen specifically from https://github.com/stellar/xdrgen/pull/163 correct?

overcat: No, I merged the following [four PRs](https://github.com/stellar/xdrgen/pulls?q=is%3Aopen+is%3Apr+author%3Aovercat+java) locally and then generated the XDR classes in this PR with the final code.

src/main/java/org/stellar/sdk/Predicate.java
```diff
@@ -186,15 +187,15 @@ public AbsBefore(TimePoint timePoint) {
     }
 
     public AbsBefore(long epochSeconds) {
-      this(new TimePoint(new Uint64(epochSeconds)));
+      this(new TimePoint(new Uint64(new XdrUnsignedHyperInteger(epochSeconds))));
     }
 
     public long getTimestampSeconds() {
-      return timePoint.getTimePoint().getUint64();
+      return timePoint.getTimePoint().getUint64().getNumber().longValue();
```
sreuland, line 194: this won't work well in cases where the value is larger than signed long can hold, coercing the biginteger into signed long has potential to force it to just the low-order of 64 bits:
example of issue - https://www.geeksforgeeks.org/biginteger-longvalue-method-in-java/
spec - https://docs.oracle.com/javase/8/docs/api/java/math/BigInteger.html#longValue--

I think the best approach is to preserve the original value in the the BigInteger and return it instead on these methods, such as:
`public BigInteger getTimestampSeconds()`

sreuland (changes requested):
the updates look fine, before merging can you respond to the comment about BigInteger conversions to longs here:

https://github.com/stellar/java-stellar-sdk/pull/498/files#r1285301854

thanks!

src/main/java/org/stellar/sdk/Predicate.java
```diff
@@ -186,15 +187,15 @@ public AbsBefore(TimePoint timePoint) {
...
     }
 
     public Instant getDate() {
-      return Instant.ofEpochSecond(timePoint.getTimePoint().getUint64());
+      return Instant.ofEpochSecond(timePoint.getTimePoint().getUint64().getNumber().longValue());
```
overcat, line 198: Hi @sreuland, do you have any suggestions for here? The maximum value that Instant.ofEpochSecond can accept is 31556889864403199. When the value exceeds this limit, should we throw a warning, or remove this method?

src/main/java/org/stellar/sdk/Predicate.java
overcat: Fixed in 8d28f047413d5d67855dcd556ef543f276c2b5ae

src/main/java/org/stellar/sdk/xdr/XdrUnsignedHyperInteger.java
sreuland: ok, as the precedent in this repo is to commit generated code, we do have to make sure the same generated code is repeatable, or we will have regressions later when anyone else tries to generate code. 

this could be the time to consider automating the codegen with Makefile rather than relying on docs in readme, for example, same as stellar/go repo does with [make xdr-update](https://github.com/stellar/go/blob/soroban-xdr-next/Makefile) . If that were in place then the version(commit hash) of xdrgen and the xdr schema `.x` files are pinned in there, and would at least not need to commit the `.x` files in this repo any more since the versions are captured as  pinned constants in the Makefile.

We could create new issue for 'create makefile for xdr code gen' to do this later, and just update readme.md for now and state how the version of xdrgen used here was created locally based on the merge of several pending pr's just for TODO and visibility.

Once the xdrgen pr's are merged, that would be a good time to consider doing the Makefile and re-run the xdrgen built from xdrgen/master to confirm the same code is produced at same time to tie it up.

src/main/java/org/stellar/sdk/Predicate.java
sreuland: ok, the `java.time` package has same limitation on signed longs for epochs. Throwing would not be great because it would result in potential sdk errors from valid data from the api server. 

The best tradeoff can think of is to check if larger than Long.MAX and if so, return Instant.MAX instead as the best 'far future' native date representation possible in java and add a javadoc to highlight the coercion:

```
/**
     * Gets the java date representation of a Predicate
     * If the Predicate specifies an epoch larger than 31556889864403199, it will
     * coerce to Instant.MAX instead as no greater value can be represented.
     *
     * @return Instant   the java date representation of the predicate
     */
    public Instant getDate() {
       ...
    }

``` 

at the same time add a new fall-back method that provides the original data so that it's never lost:

```
/**
     * Gets the Predicate epoch value
     *
     * @return BigInteger   the predicate epoch in seconds
     */
    public BigInteger getEpochSeconds() {
     return timePoint.getTimePoint().getUint64().getNumber()
    }

```

sreuland (approved):
the changes look good, one remaining discussion of overflow on datetime's from the xdr HyperIntegers  - https://github.com/stellar/java-stellar-sdk/pull/498#discussion_r1286225243

once that is resolved, we can merge?



beyond changes from that, the rest looks good.

src/main/java/org/stellar/sdk/Predicate.java
overcat: We already have `getTimestampSeconds`, and I believe it represents the same concept as `getEpochSeconds`. So, do we really need to introduce a new function?

Fixed in a68b0cb21ccf0055b03d434f24e5be469602c3c7

sreuland: no, sounds good, I did not notice `getTimestampSeconds` was already there.

I'll merge this PR shortly unless you had anything else, thanks!

Issue #502, opened 2023-08-08
Optimize the way of generating XDR classes.


overcat:
### What problem does your feature solve?
dependent on manual execution of xdrgen, this offline step may be error-prone and does not lead to reproducible builds, since the commit hash/versions used at generation time for xdrgen and the xdr files themselves are are not captured.

See https://github.com/stellar/java-stellar-sdk/pull/498#discussion_r1286197370

### What would you like to see?

Add a Makefile build script similar to this on: https://github.com/stellar/go/blob/master/Makefile

### What alternatives are there?
N/A

PR #503, opened 2023-08-09
Add the fromXdrBase64, fromXdrByteArray, toXdrBase64 and toXdrByteArray methods to the XDR classes.

### LINKED ISSUE #501: Add the fromXdrBase64, fromXdrByteArray, toXdrBase64 and toXdrByteArray methods to the XDR classes.
1. Currently, there is a large amount of template code in the current codebase. Moreover, as the project develops, similar template code may increase. I believe that we should generate corresponding methods in the XDR classes.
Some examples:
  - https://github.com/stellar/java-stellar-sdk/blob/25dc8a8fa3b38fc6a12fa212ee1b51665a113135/src/main/java/org/stellar/sdk/SorobanServer.java#L480-L516
  - https://github.com/stellar/java-stellar-sdk/blob/25dc8a8fa3b38fc6a12fa212ee1b51665a113135/src/main/java/org/stellar/sdk/Util.java#L104-L162
  - https://github.com/stellar/java-stellar-sdk/blob/25dc8a8fa3b38fc6a12fa212ee1b51665a113135/src/main/java/org/stellar/sdk/AbstractTransaction.java#L121-L133

2. When users use the getTransaction method to retrieve the result of a Soroban transaction, there are many XDRs that need to be parsed. Adding these two methods can facilitate users in parsing them.

Add the fromXdrBase64 and toXdrBase64 methods to the XDR classes.

```diff
diff --git a/src/main/java/org/stellar/sdk/xdr/AccountEntry.java b/src/main/java/org/stellar/sdk/xdr/AccountEntry.java
index 57ed904d..03be0dd1 100644
--- a/src/main/java/org/stellar/sdk/xdr/AccountEntry.java
+++ b/src/main/java/org/stellar/sdk/xdr/AccountEntry.java
@@ -6,6 +6,9 @@
 import static org.stellar.sdk.xdr.Constants.*;
 
 import com.google.common.base.Objects;
+import com.google.common.io.BaseEncoding;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.util.Arrays;
 
@@ -227,6 +230,32 @@ public boolean equals(Object object) {
         && Objects.equal(this.ext, other.ext);
   }
 
+  @Override
+  public String toXdrBase64() throws IOException {
+    BaseEncoding base64Encoding = BaseEncoding.base64();
+    return base64Encoding.encode(toXdrByteArray());
+  }
+
+  @Override
+  public byte[] toXdrByteArray() throws IOException {
+    ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+    XdrDataOutputStream xdrDataOutputStream = new XdrDataOutputStream(byteArrayOutputStream);
+    encode(xdrDataOutputStream);
+    return byteArrayOutputStream.toByteArray();
+  }
+
+  public static AccountEntry fromXdrBase64(String xdr) throws IOException {
+    BaseEncoding base64Encoding = BaseEncoding.base64();
+    byte[] bytes = base64Encoding.decode(xdr);
+    return fromXdrByteArray(bytes);
+  }
+
+  public static AccountEntry fromXdrByteArray(byte[] xdr) throws IOException {
+    ByteArrayInputStream byteArrayInputStream = new ByteArrayInputStream(xdr);
+    XdrDataInputStream xdrDataInputStream = new XdrDataInputStream(byteArrayInputStream);
+    return decode(xdrDataInputStream);
+  }
+
   public static final class Builder {
     private AccountID accountID;
     private Int64 balance;
```

N/A

overcat:
Closes #501 

The XDR class generation in this PR depends on https://github.com/stellar/xdrgen/pull/164

#### Breaking changes
- Remove `Utils.claimableBalanceIdToXDR` and `Utils.xdrToClaimableBalanceId`.


overcat:
This PR is ready for review and merge.

sreuland:
@overcat , Can you provide a brief overview or drill down on where to focus for real logic/functionality changes, key files to look at?  there's 367 changed files in the pr and trying to find efficient way to understand what's been done vs. compile-time fixes.

src/main/java/org/stellar/sdk/Util.java
```diff
@@ -103,62 +98,19 @@ public static String getSdkVersion() {
 
   public static ClaimableBalanceID claimableBalanceIdToXDR(String balanceId) {
     byte[] balanceIdBytes = BaseEncoding.base16().lowerCase().decode(balanceId.toLowerCase());
-    XdrDataInputStream balanceIdXdrDataInputStream =
-        new XdrDataInputStream(new ByteArrayInputStream(balanceIdBytes));
-
     try {
-      return ClaimableBalanceID.decode(balanceIdXdrDataInputStream);
+      return ClaimableBalanceID.fromXdrByteArray(balanceIdBytes);
```
sreuland, line 102: this seems too specific to claimablebalance to be in a generic util, and since you've added the xdr helper methods, how about refactoring this and deleting these claimable balance methods from Util and clients can just inline the one remaining line in the method for same result?

src/main/java/org/stellar/sdk/SorobanServer.java
```diff
@@ -122,8 +116,12 @@ public TransactionBuilderAccount getAccount(String accountId)
     if (entries == null || entries.isEmpty()) {
       throw new AccountNotFoundException(accountId);
     }
-    LedgerEntry.LedgerEntryData ledgerEntryData =
-        ledgerEntryDataFromXdrBase64(entries.get(0).getXdr());
+    LedgerEntry.LedgerEntryData ledgerEntryData;
+    try {
```
sreuland, line 120:
```suggestion
    try {
    LedgerEntry.LedgerEnt ledgerEntryData = LedgerEntry.LedgerEntryData.fromXdrBase64(entries.get(0).getXdr());
    
     long sequence = ledgerEntryData.getAccount().getSeqNum().getSequenceNumber().getInt64();
    return new Account(accountId, sequence);
    
    catch () {...}
```

rather than declare `ledgerEntryData` outside of try/catch, just use it inside when it's valid.

src/main/java/org/stellar/sdk/SorobanServer.java
```diff
@@ -421,7 +427,11 @@ private Transaction assembleTransaction(
       List<SorobanAuthorizationEntry> newEntries = new ArrayList<>(originalEntries);
       if (simulateHostFunctionResult.getAuth() != null) {
         for (String auth : simulateHostFunctionResult.getAuth()) {
-          newEntries.add(sorobanAuthorizationEntryFromXdrBase64(auth));
+          try {
+            newEntries.add(SorobanAuthorizationEntry.fromXdrBase64(auth));
```
sreuland, line 431: @overcat , please refer to the js soroban client for a reference on how it merges simulation results to original transaction,

https://github.com/stellar/js-soroban-client/blob/main/src/transaction.ts#L21

the java sdk should be consistent to how that handles merging auths, the js sdk will not apply simulated auths if the original tx had auths.

sreuland (changes requested):
overall looks good, just need to get response on the simulation results with auth, left comment - https://github.com/stellar/java-stellar-sdk/pull/503/files#r1289105610

can merge after that is resolved, thanks!

src/main/java/org/stellar/sdk/SorobanServer.java
overcat: I personally prefer to include as little code as possible within the try block. Assuming there is some other business logic before returning Account, should we also place it within the try block?

src/main/java/org/stellar/sdk/Util.java
overcat: Fixed in fb444200aaa92ce564c1a0a978c077becc29a423

src/main/java/org/stellar/sdk/SorobanServer.java
overcat: It seems that this is a recent change introduced in the JS SDK. Let's track it in the [new issue](https://github.com/stellar/java-stellar-sdk/issues/504).

overcat:
> The one follow-up is can you add a comment on the readme.md in the xdrgen section that states what git ref of stellar/xdrgen and stellar/stellar-xdr/next was used?

Let's track it in #502 and I'm pushing it forward.

PR #505, opened 2023-08-10
Preserve existing auth entries after simulation if they're present

### LINKED ISSUE #504: Preserve existing auth entries after simulation if they're present
See https://github.com/stellar/js-soroban-client/pull/114

> In complex auth scenarios, you may have signed entries that would be overwritten by simulation, so this just uses your existing entries. This should not affect "new" or "fresh" simulations just like before.

See https://github.com/stellar/js-soroban-client/pull/114

N/A

overcat:
Closes #504

### DIFF
--- a/src/test/java/org/stellar/sdk/SorobanServerTest.java
+++ b/src/test/java/org/stellar/sdk/SorobanServerTest.java
@@ -1098,11 +1098,7 @@ public class SorobanServerTest {
             .hostFunction(
                 ((InvokeHostFunctionOperation) transaction.getOperations()[0]).getHostFunction())
             .sourceAccount(transaction.getOperations()[0].getSourceAccount())
-            .auth(
-                Arrays.asList(
-                    auth,
-                    sorobanAuthorizationEntryFromXdrBase64(
-                        "AAAAAAAAAAAAAAABxYsr+8TwVOcyT2vyDK0+Am5Bu60abSDD19SRje0WVBEAAAAJaW5jcmVtZW50AAAAAAAAAgAAABIAAAAAAAAAAFi3xKLI8peqjz0kcSgf38zsr+SOVmMxPsGOEqc+ypihAAAAAwAAAAoAAAAA")))
+            .auth(Collections.singletonList(auth))
             .build();
     Transaction expectedTx =
         new Transaction(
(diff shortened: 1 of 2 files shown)


src/test/java/org/stellar/sdk/SorobanServerTest.java
```diff
@@ -1098,11 +1098,7 @@ public MockResponse dispatch(@NotNull RecordedRequest recordedRequest)
             .hostFunction(
                 ((InvokeHostFunctionOperation) transaction.getOperations()[0]).getHostFunction())
             .sourceAccount(transaction.getOperations()[0].getSourceAccount())
-            .auth(
-                Arrays.asList(
-                    auth,
-                    sorobanAuthorizationEntryFromXdrBase64(
-                        "AAAAAAAAAAAAAAABxYsr+8TwVOcyT2vyDK0+Am5Bu60abSDD19SRje0WVBEAAAAJaW5jcmVtZW50AAAAAAAAAgAAABIAAAAAAAAAAFi3xKLI8peqjz0kcSgf38zsr+SOVmMxPsGOEqc+ypihAAAAAwAAAAoAAAAA")))
+            .auth(Collections.singletonList(auth))
```
sreuland, line 1101: can we add one more test that validates the simulation response auth is used when no auth entries are on the prepared incoming tx?

btw, for consideration on code style, for static method refs, can lower the verboseness of it a bit with static method imports, then can refer to just the static method in code, here's how same line looks:

```
import static com.google.common.collect.ImmutableList.of;
...

.auth(of(auth))
```

sreuland (approved):
look good, left one suggestion for extra test coverage if you want.

src/test/java/org/stellar/sdk/SorobanServerTest.java
overcat: We have included such a test, which is not set up with Auth and SorobanData.
https://github.com/stellar/java-stellar-sdk/pull/505/files#diff-4085df79054bfe5baad3a9edddedc2f3fc877092ad09208fef8fcb226b808070R814


fix code style in 7a3fe1e9d7e999dfd57107b8493b79ae0c1bf44d

PR #521, opened 2023-09-05
fix the issue of unable to parse liquidity_pool_revoked effect properly.

### LINKED ISSUE #393: When the query effect data type returns 95, there will be a gson parsing exception
Hello, I found a bug in version 0.29.0. When the data type returns LiquidityPoolRevokedEffectResponse, the return parameter reserves_revoked should be accepted as an array instead of entity classes.

overcat:
fix #393 

Due to user demand for this fix, let's release it on the master branch.

~~Let's merge #522 first.~~

### DIFF
--- a/src/main/java/org/stellar/sdk/requests/LiquidityPoolsRequestBuilder.java
+++ b/src/main/java/org/stellar/sdk/requests/LiquidityPoolsRequestBuilder.java
@@ -73,17 +73,18 @@ public class LiquidityPoolsRequestBuilder extends RequestBuilder {
     uriBuilder.setQueryParameter(RESERVES_PARAMETER_NAME, String.join(",", reserves));
     return this;
   }
-  
+
   /**
    * Returns all liquidity pools the specified account is participating in.
    *
    * @param account Account ID to filter liquidity pools
    * @return current {@link LiquidityPoolsRequestBuilder} instance
-   * @see <a href="https://developers.stellar.org/api/resources/liquiditypools/list/">LiquidityPools</a>
+   * @see <a
+   *     href="https://developers.stellar.org/api/resources/liquiditypools/list/">LiquidityPools</a>
    */
   public LiquidityPoolsRequestBuilder forAccount(String account) {
-      uriBuilder.setQueryParameter(ACCOUNT_PARAMETER_NAME, account);
-      return this;
+    uriBuilder.setQueryParameter(ACCOUNT_PARAMETER_NAME, account);
+    return this;
   }
 
   /**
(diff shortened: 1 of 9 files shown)


src/main/java/org/stellar/sdk/requests/LiquidityPoolsRequestBuilder.java
overcat, line 1: This file is irrelevant to this PR, please ignore it.

src/test/java/org/stellar/sdk/requests/LiquidityPoolsRequestBuilderTest.java
overcat, line 1: This file is irrelevant to this PR, please ignore it.

sreuland: what issue or pr is it related to, this will land on master, is there a way to rebase the source branch to get cleaner change set?

sreuland:
It looks like this pr is also doing the release prep, as it has notes and version bump for 0.40.1, just to confirm, you intend to merge this and #522 to master and then we just tag release off master which would be an effective changeset of just those two, as nothing else has merged to master since 0.40.0 correct?

src/test/java/org/stellar/sdk/requests/LiquidityPoolsRequestBuilderTest.java
overcat: This is because when we merged #426, #426 was not formatted (it is a very "old" PR.). If necessary, I can create another PR to format it.

overcat:
> It looks like this pr is also doing the release prep, as it has notes and version bump for 0.40.1, just to confirm, you intend to merge this and #522 to master and then we just tag release off master which would be an effective changeset of just those two, as nothing else has merged to master since 0.40.0 correct?

Yes, let's merge these two PRs and then release it on the **master** branch. I have updated the changelog to include all the changes.

src/test/java/org/stellar/sdk/requests/LiquidityPoolsRequestBuilderTest.java
sreuland: ok, I see it mentioned in the changelog notes, which is great, thanks!

src/test/java/org/stellar/sdk/responses/EffectDeserializerTest.java
```diff
@@ -1318,4 +1321,96 @@ public void testDeserializeLiquidityPoolTradeEffect() {
         new AssetAmount(
             create("ARST:GB7TAYRUZGE6TVT7NHP5SMIZRNQA6PLM423EYISAOAP3MKYIQMVYP2JO"), "1.0000000"));
   }
+
```
sreuland, line 1324: great test coverage!

sreuland (approved):

PR #527, opened 2023-09-12
Merge branch 'master' into soroban

overcat:
This PR does not contain any code changes, it is only merging the code from the master branch into the soroban branch for us to work on removing guava.

I think we should merge this PR using the `Rebase and Merge`.


sreuland:
@overcat , to confirm, this is ready to merge as rebase, and the android-test failure is not issue?

overcat:
@sreuland Let's merge it first. I haven't found the cause of the `android test api 33` test failure yet, so let's ignore it for now. It runs fine on my local machine, but behaves inconsistently on GitHub Actions. It seems to be related to the performance of the assigned testing machines.

PR #530, opened 2023-09-16
add support for Soroban Preview 11.

### LINKED ISSUE #515: Update XDR reference
there is a newer version of xdr available for Protocol 20, soroban:
stellar-xdr:7015193879e9da7cb70d889a49eecf494fefe7e1

for reference, same update to xdr being done on mono repo here:
https://github.com/stellar/go/pull/5023/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52

the xdr generated classes in java are updated to this version of xdr and any compile time issues that may arise are resolved.

overcat:
Closes #515


overcat:
Hi @sreuland, let's increase the priority of this PR in the review queue.

tamirms:
@overcat ok I will wait to create the release until you tell me it's ready

sreuland:
@tamirms , I'm doing review on this now, I can help proceed with making release afterwards as well.

src/main/java/org/stellar/sdk/SorobanServer.java
```diff
@@ -428,6 +417,15 @@ private Transaction assembleTransaction(
     if (operation instanceof InvokeHostFunctionOperation) {
       // If the operation is an InvokeHostFunctionOperation, we need to update the auth entries if
       // existing entries are empty and the simulation result contains auth entries.
+      if (simulateTransactionResponse.getResults() == null
```
sreuland, line 420: @overcat , there was a fair amount of change done on the js client to handle the different types of simulation response payloads with different elements, as it could be for invoke, restore, and success/error, for reference the js client captured it like:
https://github.com/stellar/js-soroban-client/blob/main/src/soroban_rpc.ts#L168-L176

https://github.com/stellar/js-soroban-client/blob/main/src/transaction.ts#L111

Does the java parsing here need to be updated to same, it appears like so?

overcat: I think it would be better if we annotate these through documents. The JS client has done more parsing of the response, but we haven't, we keep the original response.

sreuland: yes, also how 'success' is parsed from the raw simulation response from rpc is different now depending on whether it's indicating a restore is need as part of state exipration, and there's the optional present of the preamble member:
https://github.com/stellar/js-soroban-client/blob/main/src/soroban_rpc.ts#L226-L231

so, I think you're suggestion for documents, means stating these json rpc aspects for now in comments for sdk users to implement their own parsing, and then implement parsing internally in the sdk for the simulation response variations in a later iteration here?

overcat: I have no intention of modifying it at the moment, as I mentioned earlier. Personally, I believe this would make things more complicated. If possible, let's keep it for now and listen to user feedback. We still have a chance to modify it before releasing the final version.

The link here contains detailed documents. Do you think I need to copy them into each field?
https://github.com/stellar/java-stellar-sdk/pull/530/files#diff-04e963815abf17aa3afc8545dd7431ba59aa861e57e3a04f53c5c33b82482305R12

https://soroban.stellar.org/api/methods/simulateTransaction#returns

sreuland (approved):
this looks good, i think the only update needed is per your consideration on additional mention in docs/comments in SimulateTransactionResponse for new aspects of the rpc json payload response differences per discussion https://github.com/stellar/java-stellar-sdk/pull/530/files#r1329008524

src/main/java/org/stellar/sdk/SorobanServer.java
sreuland: no, don't add docs on each field in SimulateTransactionResponse, I think the link to the soroban public docs is great to avoid duplicating docs here in code, and should be the source of record, but maybe we need to alert the sdk user why reading that link is needed due to the dynamic nature of the simulation response payload, how about this for the class docs on `SimulateTransactionResponse`:

```
/**
 * Response for JSON-RPC method simulateTransaction.
 *
 *  Note - The simulation response will have different model representations with different data memmbers
 *  present or absent depending representing the type of response that it is conveying. For example, the response for
 *  simulating an invoke host function, could be one of three types: error, success, or restore operation needed.
 *
 *  Please refer to the latest <a href="https://soroban.stellar.org/api/methods/simulateTransaction#returns"
 *  target="_blank">Soroban simulateTransaction documentation</a> for details on which data members of the
 *  simulation response model are keyed to each type of response.
 *
 */

```

sreuland: done - https://github.com/stellar/java-stellar-sdk/pull/530/commits/702628ca4d3438a688a5c917557baf5604c40c0b

PR #524, opened 2023-09-11
Remove the dependency on Guava.

overcat:
Part of #496 

This PR does not include XDR changes, I need to wait for the merge of https://github.com/stellar/xdrgen/pull/168 before initiating a separate XDR-related PR.

----
This PR is almost completed, but before merging it, we need to merge #525 and #527 first and remove `com.google.guava:guava:32.1.2-android` in build.gradle.

PR #534, opened 2023-09-19
Remove the dependency on Guava.

overcat:
See https://github.com/stellar/java-stellar-sdk/pull/524
See https://github.com/stellar/java-stellar-sdk/pull/524#issuecomment-1724736538


overcat:
Since #524 has already been approved, and this PR only fixes errors in resolving conflicts, I no longer need to wait for another approval. Let me merge it directly.

PR #541, opened 2023-09-21
Fix the issue where the validity of the encoded strkey is not verified in certain scenarios.

### LINKED ISSUE #540: `AccountConverter.enableMuxed()` not throwing exceptions with trailing characters of a valid mux account

`0.41.0`

### What did you do?
I am seeing a bug in the Java SDK. The AccountConverter.enableMuxed() isn't throwing the expected exception when passed a invalid mux account: `MCEO75Y6YKE53HM6N46IJYH3LK3YYFZ4QWGNUKCSSIQSH3KOAD7BEAAAAAAAAAAAPNT2W___`

Here is the code snippet.
```
package org.example;

import org.stellar.sdk.AccountConverter;

public class Main {
  public static void main(String[] args) {
    AccountConverter accountConverter = AccountConverter.enableMuxed();
    // This is a valid muxed account
    accountConverter.encode(
        "MCEO75Y6YKE53HM6N46IJYH3LK3YYFZ4QWGNUKCSSIQSH3KOAD7BEAAAAAAAAAAAPNT2W");

    // This is an invalid muxed account and we expect an IllegalArgumentException
    try {
      accountConverter.encode("MCEO75");
    } catch (IllegalArgumentException ex) {
      System.out.println("MCE075 is not valid");
      ex.printStackTrace();
    }

    // ****** PROBLEM ******
    // This is an invalid muxed account and we expect an IllegalArgumentException but nothing is
    // thrown.
    try {
      accountConverter.encode(
          "MCEO75Y6YKE53HM6N46IJYH3LK3YYFZ4QWGNUKCSSIQSH3KOAD7BEAAAAAAAAAAAPNT2W___");
    } catch (IllegalArgumentException ex) {
      System.out.println(
          "MCEO75Y6YKE53HM6N46IJYH3LK3YYFZ4QWGNUKCSSIQSH3KOAD7BEAAAAAAAAAAAPNT2W___ is not valid");
      System.out.println(ex);
    }
  }
}
```
The 3rd test should throw exception.

### What did you see instead?
No exception was thrown.

overcat:
We have replaced the base32 library, and their internal implementations are slightly different, which has caused a bug.

fix #540

### DIFF
--- a/src/main/java/org/stellar/sdk/StrKey.java
+++ b/src/main/java/org/stellar/sdk/StrKey.java
@@ -432,10 +432,8 @@ class StrKey {
   private static byte[] base32decode(byte[] data) {
     // Apache commons codec Base32 class will auto remove the illegal characters, this is
     // what we don't want, so we need to check the data before decoding
-    for (byte b : data) {
-      if (!base32Codec.isInAlphabet(b)) {
-        throw new IllegalArgumentException("Invalid Base32 character: " + (char) b);
-      }
+    if (!base32Codec.isInAlphabet(data, false)) {
+      throw new IllegalArgumentException("Invalid base32 encoded string");
     }
     return base32Codec.decode(data);
   }
--- a/src/main/java/org/stellar/sdk/StrKey.java
+++ b/src/main/java/org/stellar/sdk/StrKey.java
@@ -5,8 +5,10 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Optional;
+import org.apache.commons.codec.CodecPolicy;
 import org.apache.commons.codec.binary.Base32;
 import org.apache.commons.codec.binary.Base32OutputStream;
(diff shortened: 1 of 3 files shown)


src/main/java/org/stellar/sdk/StrKey.java
```diff
@@ -23,7 +25,8 @@ class StrKey {
 
   public static final int ACCOUNT_ID_ADDRESS_LENGTH = 56;
   private static final byte[] b32Table = decodingTable();
-  private static final Base32 base32Codec = new Base32();
+  private static final Base32 base32Codec =
```
sreuland, line 28: this doesn't seem to be necessary to configure with additional settings and STRICT, since `base32decode()` determines validity now, and it may lead to partial duplication of functionality. I tried reverting this line locally, and the tests still pass.

sreuland (approved):

overcat:
Hi @sreuland, can you check the status? I retriggered the CI, but it was still unable to merge.

PR #548, opened 2023-10-06
make `StrKey` public.

### LINKED ISSUE #318: Feature Request: Provide a simple way for decoding the public key as simple String.
Hi, Stellar SDK team

Provide a simple way for decoding the public key as simple String

`StrKey.class` is closed. Maybe makes sense to make it opened?

I don't see any other solutions. E.g. I have '_Set Option Operation_' with the _Signer_ object. I have used `KeyPair.fromXdrSignerKey(operation.signer).accountId` for retrieve String representation of Public Key. But in some cases, it returns '_Invalid Public Key_' exception -> '_not a valid GroupElement_'. Good. But I just need to try to decode the byte array in this case.
Can you please suggest any other solution to retrieve a simple String representation of Public Key e.g. from the byte array? 

On IOS SDK it retrieved like `PublicKey(xdr: x).accountId`

On Backend just used:
```
key = StrKey.decode_ed25519_public_key(public_key)
return Keypair.from_raw_ed25519_public_key(key)
```
**Summary:**  It would be nice if the SDK had simple methods available for decoding, for example, an array of bytes (e.g for _ed25519.uint256_). As in the _StrKey.class_, which is closed. 

Thanks.

overcat:
Closes #318 

In Soroban, we have many scenarios that require converting strkey into raw data. Making strkey public helps users handle it conveniently.

We will prudently make the following methods public:
- encodeEd25519PublicKey
- decodeEd25519PublicKey
- encodeEd25519SecretSeed
- decodeEd25519SecretSeed
- encodePreAuthTx
- decodePreAuthTx
- encodeSha256Hash
- decodeSha256Hash
- encodeSignedPayload
- decodeSignedPayload
- encodeContract
- decodeContract

In addition, we have modified the names of some functions to be consistent with other SDKs. Since StrKey was not previously public, there will be no destructive updates.

### DIFF
--- a/src/main/java/org/stellar/sdk/StrKey.java
+++ b/src/main/java/org/stellar/sdk/StrKey.java
@@ -18,28 +18,106 @@ import org.stellar.sdk.xdr.XdrDataInputStream;
 import org.stellar.sdk.xdr.XdrDataOutputStream;
 import org.stellar.sdk.xdr.XdrUnsignedHyperInteger;
 
-class StrKey {
+/**
+ * StrKey is a helper class that allows encoding and decoding Stellar keys to/from strings, i.e.
+ * between their binary and string (i.e. "GABCD...", etc.) representations.
+ *
+ * <p>For encoding and decoding MuxedAccount, please see {@link AccountConverter}.
+ */
+public class StrKey {
 
-  public static final int ACCOUNT_ID_ADDRESS_LENGTH = 56;
   private static final byte[] b32Table = decodingTable();
   private static final Base32 base32Codec = Base32Factory.getInstance();
 
-  public static String encodeContractId(byte[] data) {
-    char[] encoded = encodeCheck(VersionByte.CONTRACT, data);
-    return String.valueOf(encoded);
-  }
-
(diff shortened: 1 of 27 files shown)


src/main/java/org/stellar/sdk/StrKey.java
overcat, line 1: We only need to focus on `StrKey.java` and `StrKeyTest.java`, we don't need to worry about other files.

src/main/java/org/stellar/sdk/StrKey.java
```diff
@@ -428,4 +493,33 @@ private static boolean isInAlphabet(final byte[] arrayOctet) {
     }
     return true;
   }
+
```
sreuland, line 496: minor, can you pull this out to it's own file `VersionByte.java`

overcat: I think placing it here would also be a good choice, as there are no other parts relying on StrKey. Moreover, it is not public, so in the future, if necessary, we can still modify it without introducing breaking changes.

PR #549, opened 2023-10-20
Update `LedgerResponse` and `AccountResponse`, remove outdated fields, and add missing fields.

### LINKED ISSUE #287: Horizon 1.5.0 support
[Horizon 1.5.0](https://github.com/stellar/go/releases/tag/horizon-v1.5.0) added a new field to the ledgers resource called `tx_set_operation_count`, if your SDK has support for `/ledgers` or `ledger` resources, then it should include this new field. The value can be a `number` or `null`.

This value represents the total number of operations in the transactions set. You can use it to find out if recent ledgers have been full (and therefore subject to surge pricing) or not.

## Are you curious about other changes in Horizon?

My two picks are:

1. The removal of `--ingest-failed-transactions` flag, which means from now on Horizon will always ingest failed transactions. If you run your own Horizon, this might have an impact on your database.
2. A new experimental mode of doing database reingestion which uses a Stellar Core subprocess instead of a persistent Stellar Core database. Please give this a try and let us know if you find any issue!

You can see the full changelog in the release notes https://github.com/stellar/go/releases/tag/horizon-v1.5.0

### LINKED ISSUE #455: Add `AccountResponse.Balance.getIsClawbackEnabled()`
The [`AccountResponse.Balance`](https://github.com/stellar/java-stellar-sdk/blob/master/src/main/java/org/stellar/sdk/responses/AccountResponse.java#L284) object does not have an attribute that corresponds to the `is_clawback_enabled` attribute returned from Horizon.

Example [account](https://horizon-testnet.stellar.org/accounts/GA27VME5Z4AGZSAWQU3N45JUQJJPASBBBSLTFCASV7DXUMVTEJX2QD64) with `is_clawback_enabled`.

We need to add a getter function to the Java inner object to fetch this info.

overcat:
Closes #287, Closes #455


src/main/java/org/stellar/sdk/responses/AccountResponse.java
sreuland, line 1: for consideration, add reference to https://developers.stellar.org/docs/fundamentals-and-concepts/stellar-data-structures/accounts to the `AccountResponse` class docs, noticed you added similar link for `LedgerResponse` , that content looks great!

sreuland (approved):

src/main/java/org/stellar/sdk/responses/AccountResponse.java
overcat: > or consideration, add reference to https://developers.stellar.org/docs/fundamentals-and-concepts/stellar-data-structures/accounts to the AccountResponse class docs, noticed you added similar link for LedgerResponse , that content looks great!

Added.

> Also, do you feel it's worthwhile to refactor AccountResponse to use lombok also, noticed you included that refactoring on pr for LedgerResponse, understand if not, as it's not related to the change scope.

I did not refactor AccountResponse using Lombok because it would cause breaking changes. I think I will have a separate PR later to use Lombok to rewrite all the code.

PR #551, opened 2023-10-21
Add support for muxed accounts in `PaymentOperationResponse`.

### LINKED ISSUE #355: `to_muxed` and `to_muxed_id` not available in PaymentOperationResponse
0.26.0

### What did you do?
I tried accessing `to_muxed` from `PaymentOperationResponse`, but it wasn't there. For example, `PathPaymentBaseOperationResponse` has a function `getToMuxed` which returns a Muxed account from `to_muxed` and `to_muxed_id` if they are available.

I expected to be able to get the `to_muxed` property from a payment response.

### What did you see instead?

It isn't possible to get the `to_muxed` property from a payment response.

overcat:
Fix #355

### DIFF
--- a/src/main/java/org/stellar/sdk/responses/operations/PaymentOperationResponse.java
+++ b/src/main/java/org/stellar/sdk/responses/operations/PaymentOperationResponse.java
@@ -3,8 +3,11 @@ package org.stellar.sdk.responses.operations;
 import static org.stellar.sdk.Asset.create;
 
 import com.google.gson.annotations.SerializedName;
+import java.math.BigInteger;
+import java.util.Optional;
 import org.stellar.sdk.Asset;
 import org.stellar.sdk.AssetTypeNative;
+import org.stellar.sdk.responses.MuxedAccount;
 
 /**
  * Represents Payment operation response.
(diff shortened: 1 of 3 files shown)


src/main/java/org/stellar/sdk/responses/operations/PaymentOperationResponse.java
```diff
@@ -64,7 +87,21 @@ public String getFrom() {
     return from;
   }
 
+  public Optional<MuxedAccount> getFromMuxed() {
```
sreuland, line 90: 'Optional' usage in the interface seems duplicative, `null` can convey same meaning with much less code. Actualy, how do you feel about just converting PaymentOperationResponse to a `lombok` value class instead and avoid all these manuall set/get'rs?

overcat: We have similar code in [other classes](https://github.com/stellar/java-stellar-sdk/blob/19570b4abd51b11f24b3dd0c9f0881bba39d54e0/src/main/java/org/stellar/sdk/responses/operations/PathPaymentBaseOperationResponse.java#L71-L83), and I want to keep it consistent with them, so I wrote it like this.

Regarding the use of `Optional` or `null` to represent empty values, I believe that using null can indeed make our code look more concise. However, in this case, users must rely on documentation to determine whether a field may be empty. But if `Optional` is introduced, if a field is not set as an `Optional` type, then users can assume that this field definitely exists, which may be more user-friendly.

sreuland (approved):

PR #554, opened 2023-10-29
Use `Price` instead of `String` to represent prices.

### LINKED ISSUE #292: Some envelopes with buy/sell offers decoded incorrectly

overcat:
Fix #292

### DIFF
--- a/src/main/java/org/stellar/sdk/CreatePassiveSellOfferOperation.java
+++ b/src/main/java/org/stellar/sdk/CreatePassiveSellOfferOperation.java
@@ -18,13 +18,10 @@ public class CreatePassiveSellOfferOperation extends Operation {
   private final Asset selling;
   private final Asset buying;
   private final String amount;
-  private final String price;
+  private final Price price;
 
   private CreatePassiveSellOfferOperation(
-      @NonNull Asset selling,
-      @NonNull Asset buying,
-      @NonNull String amount,
-      @NonNull String price) {
+      @NonNull Asset selling, @NonNull Asset buying, @NonNull String amount, @NonNull Price price) {
     this.selling = selling;
     this.buying = buying;
     this.amount = amount;
@@ -47,7 +44,7 @@ public class CreatePassiveSellOfferOperation extends Operation {
   }
 
   /** Price of 1 unit of selling in terms of buying. */
-  public String getPrice() {
+  public Price getPrice() {
     return price;
   }
 
@@ -59,7 +56,6 @@ public class CreatePassiveSellOfferOperation extends Operation {
     Int64 amount = new Int64();
     amount.setInt64(Operation.toXdrAmount(this.amount));
     op.setAmount(amount);
-    Price price = Price.fromString(this.price);
     op.setPrice(price.toXdr());
 
     org.stellar.sdk.xdr.Operation.OperationBody body =
@@ -80,7 +76,7 @@ public class CreatePassiveSellOfferOperation extends Operation {
     private final Asset selling;
     private final Asset buying;
     private final String amount;
-    private final String price;
+    private final Price price;
 
     private String mSourceAccount;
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/stellar/sdk/CreatePassiveSellOfferOperation.java
```diff
@@ -109,13 +105,25 @@ public Builder(
...
       this.price = price;
     }
 
+    /** An alias for {@link Builder#Builder(Asset, Asset, String, Price)} */
+    public Builder(
+        @NonNull Asset selling,
+        @NonNull Asset buying,
+        @NonNull String amount,
+        @NonNull String price) {
+      this.selling = selling;
```
sreuland, line 121: for consideration, convert and delegate to the other method rather than implement locally:
```
builder(selling, buying, amount, Price.fromString(price))
```

src/main/java/org/stellar/sdk/ManageBuyOfferOperation.java
```diff
@@ -124,13 +123,25 @@ public Builder(
...
       this.price = price;
     }
 
+    /** An alias for {@link Builder#Builder(Asset, Asset, String, Price)} */
+    public Builder(
+        @NonNull Asset selling,
+        @NonNull Asset buying,
+        @NonNull String amount,
+        @NonNull String price) {
+      this.selling = selling;
```
sreuland, line 139: same overloaded constructor pattern could be used here, convert and invoke the main constructor instead:

```
this(selling, buying, amount, Price.fromString(price))

```

src/main/java/org/stellar/sdk/ManageSellOfferOperation.java
```diff
@@ -124,13 +123,25 @@ public Builder(
...
       this.price = price;
     }
 
+    /** An alias for {@link Builder#Builder(Asset, Asset, String, Price)} */
+    public Builder(
+        @NonNull Asset selling,
+        @NonNull Asset buying,
+        @NonNull String amount,
+        @NonNull String price) {
+      this.selling = selling;
```
sreuland, line 139: same as prior comment - https://github.com/stellar/java-stellar-sdk/pull/554/files#r1379165851

src/test/java/org/stellar/sdk/OperationTest.java
```diff
@@ -821,6 +879,34 @@ public void testCreatePassiveSellOfferOperation() throws IOException, FormatExce
         operation.toXdrBase64(AccountConverter.enableMuxed()));
   }
 
+  @Test
```
sreuland, line 882: nice additional coverage on this!

sreuland (approved):

PR #560, opened 2023-12-16
Modify the types of some fields in Soroban Request/Response.

overcat:
See #559 

| field                                   | before  | now  |
| --------------------------------------- | ------- | ---- |
| GetEventsRequest.startLedger            | String  | Long |
| GetEventsResponse.EventInfo.ledger      | Integer | Long |
| GetLatestLedgerResponse.protocolVersion | Integer | Long |

### DIFF
--- a/src/test/java/org/stellar/sdk/SorobanServerTest.java
+++ b/src/test/java/org/stellar/sdk/SorobanServerTest.java
@@ -555,7 +555,7 @@ public class SorobanServerTest {
             .build();
     GetEventsRequest getEventsRequest =
         GetEventsRequest.builder()
-            .startLedger("100")
+            .startLedger(100L)
             .filter(eventFilter)
             .pagination(paginationOptions)
             .build();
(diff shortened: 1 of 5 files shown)


src/test/java/org/stellar/sdk/SorobanServerTest.java
```diff
@@ -555,7 +555,7 @@ public void testGetEvents() throws IOException, SorobanRpcErrorResponse {
             .build();
     GetEventsRequest getEventsRequest =
         GetEventsRequest.builder()
-            .startLedger("100")
+            .startLedger(100L)
```
sreuland, line 558: could you add an assertion on `GetEventsResponse.EventInfo.ledger` marshaling correctly as a number now from raw json/rpc payload also?

sreuland (approved):
lgtm, left one suggestion on test coverage.

src/test/java/org/stellar/sdk/SorobanServerTest.java
overcat: We already have such an assertion. https://github.com/stellar/java-stellar-sdk/blob/3d72a9b78bb053ce016476a841b35acf43d5f37a/src/test/java/org/stellar/sdk/SorobanServerTest.java#L591

PR #565, opened 2024-01-12
Remove deprecated classes and methods.

overcat:
The following classes and methods have been marked as deprecated in previous releases, and now they have been removed
  * `AccountResponse.Signer#getAccountId()` has been removed, use `AccountResponse.Signer#getKey()` instead.
  * `OffersRequestBuilder#forAccount(String)` has been removed, use `OffersRequestBuilder#forSeller(String)` instead.
  * `RootResponse#getProtocolVersion()` has been removed, use `RootResponse#getCurrentProtocolVersion()` instead.
  * `SetOptionsOperationResponse#getSigner()` has been removed, use `SetOptionsOperationResponse#getSignerKey()` instead.
  * `Transaction.Builder` has been removed, use `TransactionBuilder` instead.
  * `TransactionBuilder#buildTimeBounds(long, long)` has been removed, use `TimeBounds#TimeBounds(long, long)` instead.
  * `TransactionBuilder#addTimeBounds(TimeBounds)` has been removed, use `TransactionBuilder#addPreconditions(TransactionPreconditions)` instead.

### DIFF
--- a/src/main/java/org/stellar/sdk/TransactionBuilder.java
+++ b/src/main/java/org/stellar/sdk/TransactionBuilder.java
@@ -8,7 +8,6 @@ import java.util.Collection;
 import java.util.List;
 import java.util.function.Function;
 import lombok.NonNull;
-import org.stellar.sdk.TransactionPreconditions.TransactionPreconditionsBuilder;
 import org.stellar.sdk.xdr.SorobanTransactionData;
 
 /** Builds a new Transaction object. */
@@ -19,8 +18,9 @@ public class TransactionBuilder {
   private final List<Operation> operations;
   private Long baseFee;
   private final Network network;
-  private TransactionPreconditions preconditions;
+  @NonNull private TransactionPreconditions preconditions;
   private SorobanTransactionData sorobanData;
+  private BigInteger txTimeout;
 
   /**
    * Construct a new transaction builder.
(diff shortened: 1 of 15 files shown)


src/main/java/org/stellar/sdk/TransactionBuilder.java
```diff
@@ -151,11 +129,7 @@ public TransactionBuilder addTimeBounds(@NonNull TimeBounds timeBounds) {
    * @param timeout Timeout in seconds.
    * @return updated Builder
    * @see TimeBounds
-   * @deprecated this method will be removed in upcoming releases, use <code>addPreconditions()
```
overcat, line 154: It is worth noting separately that I did not remove this function because it is one of the most commonly used functions in this class. I prefer to keep it for easier access. If you have a different opinion, please let me know.

sreuland: Understood, to maintain this, you can see the methods have to implement that unfortunate check/throw on the internal preconditions anyways which could be confusing for users, and retaining this leads to clutter in the sdk interface with multiple ways for users to set time bounds on a tx rather than a single approach. I would opt for less code and removing this and nudging sdk users to use preconditions instead since the underlying xdr was modeled to that, but defer to your preference as you noted for sdk ux.

sreuland (approved):
lgtm, I left one comment for consideration on time bounds - https://github.com/stellar/java-stellar-sdk/pull/565/files#r1450823070

also, do you think it's worth capturing `Breaking change` liner notes to CHANELOG on this pr, itemizes each removed method and the recommended method to use in place? since the deprecation comment is gone?

src/main/java/org/stellar/sdk/TransactionBuilder.java
overcat: I agree with removing `TransactionBuilder#addTimeBounds(TimeBounds)`, but from the user's perspective, I still hope to keep `setTimeout`. I made some modifications in ca73373ae15554fbb0f4188beb0469aa2ca28bac so that we can both keep it and possibly simplify the code to make it look less complicated. If this looks good, I will merge it.

sreuland: ok, sounds good.

PR #566, opened 2024-01-12
Fix the `hashCode` and `equals` methods in `Transaction` and `FeeBumpTransaction`.

overcat:
Fix the `hashCode` and `equals` methods in `Transaction` and `FeeBumpTransaction`, now they will compare based on the final generated XDR objects. If the XDR objects are the same, then the original instances are considered equal.

### DIFF
--- a/src/main/java/org/stellar/sdk/AbstractTransaction.java
+++ b/src/main/java/org/stellar/sdk/AbstractTransaction.java
@@ -5,6 +5,8 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
 import lombok.NonNull;
 import org.stellar.sdk.xdr.DecoratedSignature;
 import org.stellar.sdk.xdr.Hash;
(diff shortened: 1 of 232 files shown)


src/main/java/org/stellar/sdk/AbstractTransaction.java
```diff
@@ -15,8 +15,7 @@
 import org.stellar.sdk.xdr.TransactionSignaturePayload;
 
 /** Abstract class for transaction classes. */
-@EqualsAndHashCode(exclude = {"accountConverter"})
-// TODO: maybe we should not exclude accountConverter from equals and hashCode
+@EqualsAndHashCode
```
sreuland, line 18: this will introduce non-deterministic equality or false mis-matches on equals/hash? since AccountConverter is a stateless function not a value object, it can produce the same output account values whether it's `enableMuxed` property is true or false when the incoming accounts of the tx are edd25519, this would result in the a sub-class object instance with AccountConverter(enabledMuxed=false) to not match based on equals/hash to another instance of same sub-class but with AccountConverter(enabledMuxed=true), even though both of those instances will result in generated the exact same tx xdr via `toXdr()`. 

I'm wondering if this abstract tx class should even have an equals/hash code implementation at all, b/c the subclasses of this should decide how tx instance state is created for equals/hash. 

Further, it seems like lombok `@EqualsAndHashCode` which derives instance equality from member value comparisons alone is not well suited for Transaction sub classes(Transaction and FeeBumpTransaction) equality, since subclass instance state does not represent the complete value of a stellar transaction yet, correct? What do you think about the equals/hash of Transaction and FeeBumpTransaction using the output of `toXdr()` as instance value instead of member values? since that compiled xdr value represents an instance's true stellar transaction?

overcat: Thank you for pointing this out, comparing XDR objects is indeed a good approach, and I have already made the modification.

src/main/java/org/stellar/sdk/FeeBumpTransaction.java
```diff
@@ -202,4 +201,21 @@ public FeeBumpTransaction build() {
...
+  @Override
+  public boolean equals(Object object) {
+    if (this == object) {
+      return true;
+    }
+    if (object == null || getClass() != object.getClass()) {
+      return false;
+    }
+    FeeBumpTransaction that = (FeeBumpTransaction) object;
+    return Objects.equals(toEnvelopeXdr(), that.toEnvelopeXdr());
```
sreuland, line 214: how about using the `hash()` method from AbstractTransaction instead of `toEnvelopeXdr()` it encapsulates same goal of serialization of tx envelope, but also compiles in the network passphrase to the hash.

src/main/java/org/stellar/sdk/FeeBumpTransaction.java
```diff
@@ -202,4 +201,21 @@ public FeeBumpTransaction build() {
...
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(toEnvelopeXdr());
```
sreuland, line 219: same question here, can the `hash()` method from AbstractTransaction be used instead?

overcat: now it based on `signatureBase()`.

sreuland (approved):
lgtm, thanks for discussion on tx equals/hash

PR #567, opened 2024-01-12
Add `TransactionBuilder#TransactionBuilder(Transaction)` constructor.

### LINKED ISSUE #456: Signers were made immutable causing a breaking change

Java 0.29.0 upgrading to 0.37.2

### What did you do?

`transaction.getSignatures().clear();`

With older versions of the SDK, I could clear the signatures and perform my necessary logic.

 - Could we get a `AbstractTransaction.clearSignatures()` method added to the class?
 - Is there an alternate way to clear signatures?

### What did you see instead?

The signatures list is now immutable. Our engineer has to arcanely copy and modify code from the SDK in order to remove the signature; which is not the right solution.

This PR https://github.com/stellar/java-stellar-sdk/issues/410 made the signers list immutable.

overcat:
Closes #456

See https://github.com/stellar/java-stellar-sdk/issues/456#issuecomment-1257291282
> Other option is to rebuild the tx with TransactionBuilder, and then re-invoke the sign() methods to build out the signatures again.

### DIFF
--- a/src/main/java/org/stellar/sdk/TransactionBuilder.java
+++ b/src/main/java/org/stellar/sdk/TransactionBuilder.java
@@ -2,8 +2,10 @@ package org.stellar.sdk;
 
 import static org.stellar.sdk.TransactionPreconditions.TIMEOUT_INFINITE;
 
+import java.io.IOException;
 import java.math.BigInteger;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.function.Function;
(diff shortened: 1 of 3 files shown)


src/main/java/org/stellar/sdk/TransactionBuilder.java
```diff
@@ -57,6 +59,41 @@ public TransactionBuilder(TransactionBuilderAccount sourceAccount, Network netwo
     this(AccountConverter.enableMuxed(), sourceAccount, network);
   }
 
+  /**
+   * Construct a new transaction builder from a transaction.
+   *
+   * @param transaction the transaction to rebuild
+   */
+  public TransactionBuilder(Transaction transaction) {
```
sreuland, line 67: nice, this builder pattern is good way to maintain immutability and convenience!

sreuland (approved):

PR #570, opened 2024-01-31
Update `org.stellar.sdk.responses`, add missing fields.

overcat:
This PR updates `org.stellar.sdk.responses` to add missing fields and adjusts the order of some fields to correspond with the order in [stellar/go](https://github.com/stellar/go), making it easier for future maintenance.


overcat:
Hi @sreuland @tamirms, could you add an approving review?

src/main/java/org/stellar/sdk/responses/AccountResponse.java
```diff
@@ -26,21 +26,24 @@
...
+  @SerializedName("id")
+  private String id;
+
   @SerializedName("account_id")
   private String accountId;
 
   @SerializedName("sequence")
   private Long sequenceNumber;
 
-  @SerializedName("subentry_count")
```
sreuland, line 35: should we mention this removal in changelog for potential breaking aspect?

src/main/java/org/stellar/sdk/responses/AssetResponse.java
```diff
@@ -24,15 +24,31 @@ public class AssetResponse extends Response implements Pageable {
...
+  @SerializedName("num_liquidity_pools")
+  int numLiquidityPools;
+
+  @SerializedName("num_contracts")
+  int numContracts;
+
+  @SerializedName("num_archived_contracts")
+  int numArchivedContracts;
+
+  // TODO: remove the above fields when Horizon 3.0 is released.
```
sreuland, line 44: not sure what 3.0 will look like, maybe can remove this, tackle it later when it's known?

src/main/java/org/stellar/sdk/responses/RootResponse.java
```diff
@@ -33,6 +39,88 @@ public class RootResponse extends Response {
   @SerializedName("current_protocol_version")
   int currentProtocolVersion;
 
+  @SerializedName("supported_protocol_version")
+  Long supportedProtocolVersion;
+
   @SerializedName("core_supported_protocol_version")
   int coreSupportedProtocolVersion;
+
+  @SerializedName("_links")
```
sreuland, line 48: for consideration, `Links` stands on it's own, could be worth pulling up to it's own class file rather than static under RootResponse.

sreuland (approved):
lgtm, nice work!

src/main/java/org/stellar/sdk/responses/AssetResponse.java
overcat: I saw that [relevant comment](https://github.com/stellar/go/blob/master/protocols/horizon/main.go#L171-L177) has been added in stellar/go. But we can remove this TODO first, and then remove it here after horizon confirms its removal.

src/main/java/org/stellar/sdk/responses/RootResponse.java
overcat: Many Response models have their own Links, and I think it's easier to maintain them by putting them together, just like what we did in stellar/go: https://github.com/stellar/go/blob/master/protocols/horizon/main.go#L225-L231

PR #573, opened 2024-02-27
chore: add codecov related configuration.

overcat:
Add codecov related configuration to display test coverage on the GitHub PR page.


sreuland:
> Hi @sreuland @tamirms, could you add an approving review?

Hello @overcat, this looks great, i think we want to first check with our ops, and see if there is any preference from them on a preferred gh coverage tool, will respond with findings, thanks!

cc: @mollykarcher

sreuland:
> Hello @overcat, this looks great, i think we want to first check with our ops, and see if there is any preference from them on a preferred gh coverage tool, will respond with findings, thanks!

I spoke to our Ops team, and they recommend not using the GHA coverage tools, instead each repo is encouraged to use a code coverage tool locally in the projects build, so in java's case, can we look into using a gradle based coverage tool that can be locally run as part of build lifecycle locally, like [JaCoCo](https://reflectoring.io/jacoco/) ?

overcat:
> > Hello @overcat, this looks great, i think we want to first check with our ops, and see if there is any preference from them on a preferred gh coverage tool, will respond with findings, thanks!
> 
> I spoke to our Ops team, and they recommend not using the GHA coverage tools, instead each repo is encouraged to use a code coverage tool locally in the projects build, so in java's case, can we look into using a gradle based coverage tool that can be locally run as part of build lifecycle locally, like [JaCoCo](https://reflectoring.io/jacoco/) ?

We are using Jacoco, but I currently do not have a mandatory requirement for tests to fail when the coverage decreases, and I have added an extra step to upload the results to the Codecov platform so that we can see online which code has been tested.
https://github.com/stellar/java-stellar-sdk/blob/14a4e6e5d21b483b8321b578306d643cb498f9d2/build.gradle.kts#L131-L133

PR #650, opened 2024-10-29
fix: fix SSEStream unpredictably throws unwanted IOException("Canceled")

### LINKED ISSUE #649: SSEStream unpredictably throws unwanted IOException("Canceled")
I encountered an unpredictable and unwanted exception thrown from the SSEStream of the Java Horizon SDK (https://github.com/lightsail-network/java-stellar-sdk). 

```
java.io.IOException: Canceled
    at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.kt:72)
    at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
    at org.stellar.sdk.requests.ClientIdentificationInterceptor.intercept(ClientIdentificationInterceptor.java:22)
    at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
    at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.kt:201)
    at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:517)
    at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
    ...
```

**Here is the root cause:**

When `reconnectTimeout` expires, [restart](https://github.com/lightsail-network/java-stellar-sdk/blob/master/src/main/java/org/stellar/sdk/requests/SSEStream.java#L96) is called
eventSource.cancel() is triggered.

A new `StellarEventSourceListener` [is created for the new event source.](https://github.com/lightsail-network/java-stellar-sdk/blob/master/src/main/java/org/stellar/sdk/requests/SSEStream.java#L185)

The previously canceled `StellarEventSourceListener` (the one that we don't track) can receive the `IOException (Canceled)`  because the call was canceled in [okhttp](https://github.com/square/okhttp/blob/master/okhttp-sse/src/main/kotlin/okhttp3/sse/internal/RealEventSource.kt#L114). The origin of the exception is[ here.](https://github.com/square/okhttp/blob/master/okhttp-sse/src/main/kotlin/okhttp3/sse/internal/RealEventSource.kt#L84)

I suggest the `SSEStream` to not fire the `onFailure()` to the listener if the exception was targeting the canceled `StellarEventSourceListener`.

overcat:
fix #649

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ As this project is pre 1.0, breaking changes may happen for minor version bumps.
 
 ## Pending
 
+## 0.44.1
+### Update
+fix: fix `SSEStream` unpredictably throws unwanted `IOException("Canceled")`. ([#650](https://github.com/lightsail-network/java-stellar-sdk/pull/650))
+
 ## 0.44.0
 ### Update
 - feat: add support for Soroban-RPC v21. ([#593](https://github.com/lightsail-network/java-stellar-sdk/pull/593))
--- a/android_test/app/build.gradle.kts
+++ b/android_test/app/build.gradle.kts
@@ -68,7 +68,7 @@ dependencies {
     implementation("androidx.compose.material3:material3")
     // Since we are adding local jar(libs/stellar-sdk.jar) as dependency,
     // gradle cannot automatically download the required third-party dependencies.
-    implementation(files("libs/stellar-sdk-0.44.0.jar"))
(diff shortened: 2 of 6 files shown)


overcat:
Hi @lijamie98, this fix will be included in 0.44.1.

PR #677, opened 2025-03-03
feat: add functions to send assets to contract address without relying on Stellar RPC.

overcat:


overcat:
https://github.com/stellar/quickstart/issues/667

PR #682, opened 2025-03-17
Fix the build.gradle.kts where the version `master-SNAPSHOT:uber` was not found

lijamie98:
We should either fix the version to the latest release or create the `master-SNAPSHOT:uber`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,10 @@
 
 ## Pending
 
+## 1.2.1
+
 ### Update:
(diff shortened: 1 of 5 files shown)


overcat:
Hi @lijamie98, when the user first retrieves this library from JitPack, JitPack will automatically create a release, which takes some time. I noticed that the master uber has been created correctly. https://jitpack.io/com/github/lightsail-network/java-stellar-sdk/master-SNAPSHOT/

overcat:
But I think we can make some improvements by building immediately after each commit, instead of waiting for the user to try to fetch it: https://docs.jitpack.io/building/#building-ahead-of-time

PR #691, opened 2025-05-22
Add getSacBalance to SorobanServer

lijamie98:


src/main/java/org/stellar/sdk/SorobanServer.java
```diff
@@ -560,6 +544,52 @@ public SendTransactionResponse sendTransaction(Transaction transaction) {
         "sendTransaction", params, new TypeToken<SorobanRpcResponse<SendTransactionResponse>>() {});
   }
 
+  public GetSacBalanceResponse getSacBalance(String contractId, Asset asset, Network network)
```
lijamie98, line 547: I am not sure if creating the `GetSacBalanceRequest` class would be better because it is translated to `GetLedgerEntriesRequest`.

src/main/java/org/stellar/sdk/SorobanServer.java
```diff
@@ -560,6 +544,52 @@ public SendTransactionResponse sendTransaction(Transaction transaction) {
...
+      throws IOException {
+
+    if (!StrKey.isValidContract(contractId)) {
+      throw new IllegalArgumentException("expected contract ID, got " + contractId);
```
overcat, line 551:
```suggestion
      throw new IllegalArgumentException("Invalid contract ID: " + contractId);
```

src/main/java/org/stellar/sdk/SorobanServer.java
```diff
@@ -560,6 +544,52 @@ public SendTransactionResponse sendTransaction(Transaction transaction) {
...
+    GetLedgerEntriesResponse response = this.getLedgerEntries(Collections.singleton(ledgerKey));
+
+    List<GetLedgerEntriesResponse.LedgerEntryResult> entries = response.getEntries();
+    if (entries == null || entries.isEmpty()) {
+      return null;
+    }
+
+    GetLedgerEntriesResponse.LedgerEntryResult entry = entries.get(0);
+    LedgerEntry.LedgerEntryData ledgerEntryData =
+        LedgerEntry.LedgerEntryData.fromXdrBase64(entry.getXdr());
```
overcat, line 576: Please use `Utils.parseXdr`, and do not throw `IOException` in the function signature.

src/main/java/org/stellar/sdk/responses/sorobanrpc/GetSacBalanceResponse.java
```diff
@@ -0,0 +1,21 @@
+package org.stellar.sdk.responses.sorobanrpc;
+
+import lombok.Builder;
+import lombok.Value;
+
+@Builder
+@Value
+public class GetSacBalanceResponse {
```
overcat, line 8:
```suggestion
public class GetSACBalanceResponse {
```

src/main/java/org/stellar/sdk/responses/sorobanrpc/GetSacBalanceResponse.java
```diff
@@ -0,0 +1,21 @@
...
+  long latestLedger;
```
overcat, line 9:
```suggestion
  Long latestLedger;
```

src/main/java/org/stellar/sdk/SorobanServer.java
```diff
@@ -5,12 +5,7 @@
 import java.io.Closeable;
 import java.io.IOException;
 import java.net.SocketTimeoutException;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
-import java.util.Optional;
-import java.util.UUID;
+import java.util.*;
```
overcat, line 8: Please avoid using wildcard imports. I know both have their advantages, but this has been discussed in other PRs before. Since we have a standard now, let's adhere to it.
https://github.com/lightsail-network/java-stellar-sdk/blob/master/CONTRIBUTING.md#code-style

src/main/java/org/stellar/sdk/SorobanServer.java
overcat, line 547:
```suggestion
  public GetSACBalanceResponse getSACBalance(String contractId, Asset asset, Network network)
```

overcat, line 547: Please add documentation for this function.

overcat (changes requested):
👍 It looks great. I have left some minor suggestions for changes.

src/main/java/org/stellar/sdk/SorobanServer.java
overcat: I think it's fine just the way it is now, no need to add anything.

src/main/java/org/stellar/sdk/SorobanServer.java
lijamie98: This is actually from the auto formatter, which is google-java-formatter. What do you think we let the formatter handle this?

src/main/java/org/stellar/sdk/SorobanServer.java
```diff
@@ -560,6 +544,52 @@ public SendTransactionResponse sendTransaction(Transaction transaction) {
...
+                            Arrays.asList(Scv.toSymbol("Balance"), Scv.toAddress(contractId))))
+                    .durability(ContractDataDurability.PERSISTENT)
+                    .build())
+            .build();
+
+    GetLedgerEntriesResponse response = this.getLedgerEntries(Collections.singleton(ledgerKey));
+
+    List<GetLedgerEntriesResponse.LedgerEntryResult> entries = response.getEntries();
+    if (entries == null || entries.isEmpty()) {
+      return null;
```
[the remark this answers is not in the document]
lijamie98: I changed it to the same as the JS SDK.

src/main/java/org/stellar/sdk/SorobanServer.java
overcat: I don't think google-java-formatter does this; are you using IDEA? IDEA does this by default.
https://www.baeldung.com/intellij-disable-wildcard-import

overcat: Sorry, I just tested it, and it was really modified by google-java-formatter.

PR #692, opened 2025-05-23
Add `toBigInteger()` to 128-bit and 256-bit integers

lijamie98:
In Anchor Platform, we have to convert int128 to BigInteger. It would be nice for the SDK to support these functions as well for convenience. 

I did not find any helper functions. Let me know if these functions already exist.

### DIFF
--- a/src/main/java/org/stellar/sdk/xdr/Int128Parts.java
+++ b/src/main/java/org/stellar/sdk/xdr/Int128Parts.java
@@ -5,6 +5,7 @@ package org.stellar.sdk.xdr;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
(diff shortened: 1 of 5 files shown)


lijamie98:
> Hi @lijamie98, you can take a look at a series of classes under the [`scval`](https://github.com/lightsail-network/java-stellar-sdk/blob/master/src/main/java/org/stellar/sdk/scval) package, such as [`ScvUint128`](https://github.com/lightsail-network/java-stellar-sdk/blob/master/src/main/java/org/stellar/sdk/scval/ScvUint128.java). Does this meet your needs?

Yes that works. Can we make the class public?

lijamie98:
oh, I can use `Scv.fromInt128()`

